PR #235, opened 2023-08-09
feat: carousel component

schirrel:
## Infos

[Task](https://juntossomosmais.monday.com/boards/3035294429/views/95907691/pulses/4758205321)

#### What is being delivered?

**Carousel**
- Atom Carousel component using Swiper
- Atom Carousel Item components
- ADR about Carousel, Ionic and Swiper
- ADR about te problem with swiper and the conversio of atom-component-item into swiper slide

** Side Effects **
- Elevation tokens (was used on carousel)
- Border Radius missing tokens  (was used on carousel)
- Fix doc abour transition token
- Create visual ui box component to be used on Storybook

#### What impacts?

Adding carousel to Atomium DS, in order to replace current carousel component, and remove one dependency on the vue 3 migration.

#### Reversal plan

Reset commit to the previous version 02b65773dbd33eade6761dedab03eee6775d0bbe
#### Evidences

Nuxt

https://github.com/juntossomosmais/atomium/assets/6757777/26b45690-c00e-4124-b099-d00c80a53123

Vue

https://github.com/juntossomosmais/atomium/assets/6757777/e3fb8ced-cb6a-472d-8bf4-16b409a393f6

React Next

image

> Tested on misty-mountain and web-customer-typescript

Elevation Docs

https://github.com/juntossomosmais/atomium/assets/6757777/8e3c5f06-cf85-49d9-b3ea-3829c770117f

### DIFF
--- a/apps/docs/stories/architecture-decision-records/0012-why-have-a-carousel-with-swiper.stories.mdx
+++ b/apps/docs/stories/architecture-decision-records/0012-why-have-a-carousel-with-swiper.stories.mdx
@@ -25,3 +25,6 @@ The [Swiper](https://swiperjs.com/)  was the only option considered.
 ## Decision
 We have decided to use Swiper as our Carousel component library, once it match our requirements and we have a previous knowledge, doing a internal wrapper to maintain the Atomium signature.
 
+
+## Caveats
+An important point to keep in mind is the need of use the [injectStyles](https://swiperjs.com/element#injecting-styles) from Swiper if it is needed to overwrite or add new styles to inside of Swiper Element Shadow Dom.
--- a/apps/docs/stories/architecture-decision-records/0012-why-have-a-carousel-with-swiper.stories.mdx
+++ b/apps/docs/stories/architecture-decision-records/0012-why-have-a-carousel-with-swiper.stories.mdx
@@ -1,0 +1,27 @@
+import { Meta } from '@storybook/addon-docs'
+
+<Meta title="Docs/Architecture Decision Records/ADR 0012: Carousel with Swiper" />
+
+# ADR 0012: Carousel with Swipe
+
+🗓️ 2023-08 · ✍️ [@schirrel](https://github.com/schirrel), [@MayaraRMA](https://github.com/MayaraRMA), [@tassioFront](https://github.com/tassioFront) and [@felipefialho](https://twitter.com/felipefialho_)
+
+## Context
+
+We need a Carousel component with consistent layout across projects.
+
+## Decision Drivers
+
+- We need a Carousel component/library must be web component;
+- The Carousel component/library must give us the possibility to adjust styles as we need to match our Style Guide;
+- The Carousel component/library should have a good integrations with Ionic components, once it is the source of our Design System.
+
+
+## Considered Options
+As we already have the usage of the [Swiper](https://swiperjs.com/) with an ond version inside an project and Ionic itself [indicates the migration](https://ionicframework.com/docs/v6/api/slides#migration) of the same once they have discontinued the ion-slides.
+The [Swiper](https://swiperjs.com/)  was the only option considered. 
+
+
+## Decision
+We have decided to use Swiper as our Carousel component library, once it match our requirements and we have a previous knowledge, doing a internal wrapper to maintain the Atomium signature.
+
--- a/apps/docs/stories/architecture-decision-records/0013-why-using-virtual-carousel-item.stories.mdx
+++ b/apps/docs/stories/architecture-decision-records/0013-why-using-virtual-carousel-item.stories.mdx
@@ -1,0 +1,44 @@
+import { Meta } from '@storybook/addon-docs'
+
+<Meta title="Docs/Architecture Decision Records/ADR 0013: Why using a virtual carousel item" />
+
+# ADR 0013: Why using a virtual carousel item
+
+🗓️ 2023-08 · ✍️ [@schirrel](https://github.com/schirrel), [@MayaraRMA](https://github.com/MayaraRMA), [@tassioFront](https://github.com/tassioFront) and [@felipefialho](https://twitter.com/felipefialho_)
+
+## Context
+Wrapping up Swiper HTMLElements as Atomium element.
+
+## Problems
+When creating `<atom-carousel>` and its items `<atom-carousel-item>`, the HTML three place the `<atom-carousel-item>` inside `<swiper-element>`, but the swiper needs its children to be `<swiper-slide>`.
+
+Then we need to convert each `atom-carousel-item` into `swiper-slide`.
+
+## Considered Options
+To convert the inner `<atom-carousel-item>` to `<swiper-slide>`, and we have two options:
+
+- Create and `atom-carousel-item` component and inside itself creating the `<swiper-slide>` component on parent and them self deleting from parent.
+- The parent itself convert all children at once.
+
+## Decision Drivers
+
+- Complexity
+- Performance
+
+### Performance
+We had performed a test with both approach using 9 items inside the Carousel:
+- With children `atom-carousel-item` doing the conversion, it took _0.50ms_
+- With parent `atom-carousel` doing the conversion, it took  _0.20ms_
+
+
+### Complexity
+Taking in consideration the complexity of both codes, not only the parent's conversion approach was using less code but take also avoid us to create a component only to destruct itself. This also avoid the have one more file to maintain, build, test and also to be on final bundle.
+
+
+## Decision
+
+Once parent's conversion approach took less than half of conversion time and its code is less complex, we had decide to have `<atom-carousel-item>` as a _"virtual component"_ that means this component is a pattern but not a real component. The usage of `atom-carousel` requires its children to be written with `atom-carousel-item` HTMLElement in order to be converter to `swiper-slide`.
+
+Besides it is not a real component, we have add a Story about it to provide the usage of it and to indicate the `lazy` property that `<swiper-slide>` has.
+
+PS: Even tho the component is not needed, in order to work properly with JSX/React import, we had an AtomCarouselItem component but only for compatibility purpose.
--- a/apps/docs/stories/architecture-decision-records/0013-why-using-virtual-carousel-item.stories.mdx
+++ b/apps/docs/stories/architecture-decision-records/0013-why-using-virtual-carousel-item.stories.mdx
@@ -17,8 +17,8 @@ Then we need to convert each `atom-carousel-item` into `swiper-slide`.
 ## Considered Options
 To convert the inner `<atom-carousel-item>` to `<swiper-slide>`, and we have two options:
 
-- Create and `atom-carousel-item` component and inside itself creating the `<swiper-slide>` component on parent and them self deleting from parent.
-- The parent itself convert all children at once.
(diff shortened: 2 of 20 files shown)


apps/docs/stories/architecture-decision-records/0013-why-using-virtual-carousel-item.stories.mdx
```diff
@@ -0,0 +1,44 @@
...
+## Problems
+When creating `<atom-carousel>` and its items `<atom-carousel-item>`, the HTML three place the `<atom-carousel-item>` inside `<swiper-element>`, but the swiper needs its children to be `<swiper-slide>`.
+
+Then we need to convert each `atom-carousel-item` into `swiper-slide`.
+
+## Considered Options
+To convert the inner `<atom-carousel-item>` to `<swiper-slide>`, and we have two options:
+
+- Create and `atom-carousel-item` component and inside itself creating the `<swiper-slide>` component on parent and them self deleting from parent.
+- The parent itself convert all children at once.
```
tassioFront, line 21: Suggestion: better to create an order list to fit with your text: 

```
1. Create and `atom-carousel-item` component and inside itself creating the `<swiper-slide>` component on parent and them self deleting from parent.
2. The parent itself convert all children at once.
```

packages/tokens/stories/components/Box.jsx
```diff
@@ -0,0 +1,42 @@
...
+        style={{
+          borderRadius: borderRadius || 'var(--border-radius-medium)',
+          boxShadow: `var(${elevation})`,
+          border: elevation ? '' : '1px solid var(--color-neutral-light-4)',
+          display: 'flex',
+          height: 'var(--spacing-xgiant)',
+          padding: 'var(--spacing-base)',
+          width: 'var(--spacing-xgiant)',
+          font: 'var(--text-badge)',
+          fontWeight: 'var(--font-weight-regular)	',
```
denilsonrp, line 16: **suggestion:**

```suggestion
          fontWeight: 'var(--font-weight-regular)',
```

apps/docs/stories/architecture-decision-records/0013-why-using-virtual-carousel-item.stories.mdx
```diff
@@ -0,0 +1,44 @@
...
+Taking in consideration the complexity of both codes, not only the parent's conversion approach was using less code but take also avoid us to create a component only to destruct itself. This also avoid the have one more file to maintain, build, test and also to be on final bundle.
+
+
+## Decision
+
+Once parent's conversion approach took less than half of conversion time and its code is less complex, we had decide to have `<atom-carousel-item>` as a _"virtual component"_ that means this component is a pattern but not a real component. The usage of `atom-carousel` requires its children to be written with `atom-carousel-item` HTMLElement in order to be converter to `swiper-slide`.
+
+Besides it is not a real component, we have add a Story about it to provide the usage of it and to indicate the `lazy` property that `<swiper-slide>` has.
+
+PS: Even tho the component is not needed, in order to work properly with JSX/React import, we had an AtomCarouselItem component but only for compatibility purpose.
```
tassioFront, line 44: Congrats on the analysis!

apps/docs/stories/architecture-decision-records/0013-why-using-virtual-carousel-item.stories.mdx
```diff
@@ -0,0 +1,44 @@
...
+- Complexity
+- Performance
+
+### Performance
+We had performed a test with both approach using 9 items inside the Carousel:
+- With children `atom-carousel-item` doing the conversion, it took _0.50ms_
+- With parent `atom-carousel` doing the conversion, it took  _0.20ms_
+
+
+### Complexity
```
tassioFront, line 34: Question: We are considering time complexity only, does make sense to specify it?

packages/core/src/components/carousel/carousel.tsx
```diff
@@ -0,0 +1,104 @@
...
+    this.swiperEl.innerHTML = this.host.innerHTML.replace(
+      /atom-carousel-item/g,
+      'swiper-slide'
+    )
+    let idx = 0
+    while (idx < this.host.children.length) {
+      if (this.host.children[idx].outerHTML.indexOf('atom-carousel-item') > -1)
+        this.host.removeChild(this.host.children[idx])
+      else idx++
+    }
```
tassioFront, line 66: Suggestion; Another option to it would be using `for` instead: 

```TS
for (const element of this.host.children) {
  if (element.outerHTML.indexOf('atom-carousel-item') > -1)
    this.host.removeChild(element)
}

// or even


for (let idx=0;idx < this.host.children.length;idx++) {
  if (this.host.children[idx].outerHTML.indexOf('atom-carousel-item') > -1)
    this.host.removeChild(this.host.children[idx])
}
```


Totally up to you, but guess it works as well

apps/docs/stories/architecture-decision-records/0013-why-using-virtual-carousel-item.stories.mdx
schirrel: Basically only @tassioFront, look for O complexity in both cases is the same, one approach the loop is on the parent, other approach has no explicity loop, but he operation repeats the same amount as children (inside of each).
The other operations are simple 1.

Besides on the parent we just have a simple string Mmanagement  and remove child.  On the children we must create a structured component, append to parent and delete the old.

Do you see that this level of details is needed?

packages/tokens/stories/components/Box.jsx
schirrel: thanks

packages/core/src/components/carousel/carousel.tsx
schirrel: @tassioFront i confess i havent even tried for. Several years ago i have faced big trouble with array manipulation/deletion inside of the array loop iself, like on the first example. Since then ialways go the to safe path 😅 

Besides that, In fact the while loop is a bit of different from the for, see:
On the while code i delete the children and the index keep the same, so when i delete array[2], the next array[3] becames the new array[2], but i keep deletin the array[2] because my idx havent change.
The for otherwise increments and jump to next, "forgeting" that the previously 3 now became 2, and then goes straight to the array[3] to delete it, this way the previously 3 that became the new 2  is never getting deleted.

Least but no last you may ask why i have the valitions with else, this is because the way component renders.
<atom-component> will have to kind of children <atom-component-item> the one we will remove from here and only existe so we get the html to be used in referente, and the correct <swiper-container>, whitout that if/else we would remove the swiper-container too.

Does this make sense to you?

apps/docs/stories/architecture-decision-records/0013-why-using-virtual-carousel-item.stories.mdx
tassioFront: Sorry, I misunderstood the text order. I saw the time comparison time, but it is in the `Performance` block, not the `Complexity` one. I don't need it is required. Keep going, my friend

packages/core/src/components/carousel/carousel.tsx
tassioFront: That makes all sense! Thanks for the clear explanation!

packages/core/src/components/carousel/carousel.tsx
```diff
@@ -0,0 +1,104 @@
...
+  @Prop({ mutable: true })
+  slidesPerView?: number | string = 1
+  @Prop({ mutable: true })
+  spaceBetween?: number = 0
+  @Prop({ mutable: true }) speed?: number
+
+  // injectStyles is needed in order to overwrite css by class inside shadow root with ::part
+  private injectStyles = ['.swiper-button-disabled { opacity: 0 !important}']
+
+  private convertChildren() {
```
tassioFront, line 56: Question: Does make sense to add the ADR link here?

schirrel: @tassioFront about the stylesInject? I confess i havent saw this as an architectural desicion once this is the way swiper  docs indicate to overwrite or add styles to shadow DOM scope, [is in their docs](https://swiperjs.com/element#injecting-styles), was not our decision in fact.

But as we already have an ADR about carousel and the swiper i can add this point there as a caveat, what you think?

tassioFront: > But as we already have an ADR about carousel and the swiper i can add this point there as a caveat, what you think?

I'm talking about this one, actually @schirrel  hahahaha Add de ADR link of Carousel and swiper at `convertChildren` function, thinking people might not understand this function at the first touch.

schirrel: Ooooh i got now 🤣  makes total sense, i think i forgot to add here and added only to carousel item 😅

schirrel: done [3983924](https://github.com/juntossomosmais/atomium/pull/235/commits/3983924369a620a702d271b83506631316ec4eff)

apps/docs/stories/architecture-decision-records/0013-why-using-virtual-carousel-item.stories.mdx
schirrel: Nice @tassioFront thanks, i've add at this commit [0287647](https://github.com/juntossomosmais/atomium/pull/235/commits/028764756a3ac42e40cf4e9952b7ed1b8daf1e2f)

packages/core/src/components/carousel/stories/carousel/carousel.core.stories.tsx
```diff
@@ -0,0 +1,98 @@
+import { Meta, StoryObj } from '@storybook/web-components'
+
+import { CarouselStoryArgs } from './carouse.args'
+import './style.css'
+export default {
+  title: 'Components/Carousel',
+  ...CarouselStoryArgs,
+} as Meta
+const createComponent = (args, itemClass: string) => {
```
igorwessel, line 9: **praise:** ‏:rocket:

igorwessel (dismissed):

apps/docs/stories/architecture-decision-records/0012-why-have-a-carousel-with-swiper.stories.mdx
```diff
@@ -0,0 +1,30 @@
+import { Meta } from '@storybook/addon-docs'
+
+<Meta title="Docs/Architecture Decision Records/ADR 0012: Carousel with Swiper" />
+
+# ADR 0012: Carousel with Swipe
```
mauriciomutte, line 5: typo:
```suggestion
# ADR 0012: Carousel with Swiper
```

mauriciomutte (dismissed):
🚀

apps/docs/stories/architecture-decision-records/0012-why-have-a-carousel-with-swiper.stories.mdx
schirrel: ops, thanks @mauriciomutte

packages/core/src/components/carousel/carousel.tsx
tassioFront: hahahahahaha perfect, my friend

tassioFront (dismissed):

denilsonrp (dismissed):

igorwessel (approved):

tassioFront (approved):

mauriciomutte (approved):

Issue #262, opened 2023-09-14
BUG - Component inside carousel item is not interactible or keep it usual behavior


schirrel:
While implementing carousing within Loja Virtual, we face the issue of the Card componing looses its "connectin" with the framework. Stops to be interactible and reactive, stops to update values and other items, see the video below:

https://github.com/juntossomosmais/atomium/assets/6757777/e967d598-320f-4378-acf0-553ec786085b

PS. I already uunderstood whats is going on and am current working on it, on the next hours i will open the PR fixint it 🙏🏻

schirrel:
merged https://github.com/juntossomosmais/atomium/pull/263

PR #263, opened 2023-09-14
fix: carousel convertion

schirrel:
## Infos

[Task](https://juntossomosmais.monday.com/boards/3035294429/views/95907691/pulses/5094521790)
https://github.com/juntossomosmais/atomium/issues/262

#### What is being delivered?

Change transform atom-carousel-item to swiper-slide approach to the second approach, once we descover the first approach breaks the usage with frameworks

- update adr 0013 explaining
- update unit tests
- create atom-carousel-item test

#### What impacts?

Allow carrousel to be used with complex framework components

#### Reversal plan

Describe which plan we should follow if this delivery  has to be reversed.

#### Evidences
Before

https://github.com/juntossomosmais/atomium/assets/6757777/e967d598-320f-4378-acf0-553ec786085b

After fix

https://github.com/juntossomosmais/atomium/assets/6757777/2c37f539-beb1-4cca-b566-7d0701fdc1c2


gabrielduete (dismissed):

apps/docs/stories/architecture-decision-records/0013-why-do-we-need-to-convert-carousel-item.stories.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+## Decision Drivers
+
+- Performance
+- Complexity
+- Compatibility
+
+### Performance
+
+We had performed a test with both approach using 9 items inside the Carousel:
```
felipefialho, line 34: ❤️

apps/docs/stories/architecture-decision-records/0013-why-do-we-need-to-convert-carousel-item.stories.mdx
```diff
@@ -0,0 +1,80 @@
...
+parentSwiperContainer.appendChild(swiperSlide)
+this.host.parentElement.removeChild(this.host)
+```
+
+Taking in consideration the complexity of both codes, not only the parent's conversion approach was using less code but take also avoid us to create a component only to destruct itself. This also avoid the have one more file to maintain, build, test and also to be on final bundle.
+
+### Compatibility
+
+This subject was in fact the decision maker, even tho the performance of the first approach took half of the time, when it came to compatibility we discovered why.
+For simples cases, such static card, plain text or image showcases, no difference between both, but when we started to use with complex components, that had to deal with framework functions such reactivity we face a bigger issue. The HTMLString conversion component loose _"connections "_ with the inner content. So it just became a plain HTML String copy, instead of a state and object reference.
```
felipefialho, line 74:
```suggestion

For simples cases, such static card, plain text or image showcases, no difference between both, but when we started to use with complex components, that had to deal with framework functions such reactivity we face a bigger issue. The HTMLString conversion component loose _"connections "_ with the inner content. So it just became a plain HTML String copy, instead of a state and object reference.
```

apps/docs/stories/architecture-decision-records/0013-why-do-we-need-to-convert-carousel-item.stories.mdx
```diff
@@ -0,0 +1,80 @@
...
+When using `.innerHTML.replace` our components lost its ability to re-render once loading is ready, update values and even trigger framework events.
+When using `appendChild` behavior, we copy the whole HTMLElement and its props, functions, children, and every memory reference we had. And working as needed and expected.
```
felipefialho, line 76:
```suggestion

When using `appendChild` behavior, we copy the whole HTMLElement and its props, functions, children, and every memory reference we had. And working as needed and expected.
```

packages/core/src/components/carousel/carousel-item/carousel-item.tsx
```diff
@@ -0,0 +1,29 @@
...
+  shadow: false,
+})
+
+export class AtomCarouselItem {
+  @Prop({ mutable: true })
+  lazy?: boolean
+
+  @Element() host!: HTMLElement
+
+  // If you need more info please read the ADR 0013
```
felipefialho, line 13: (suggestion) maybe we could link it?

packages/core/src/components/carousel/carousel-item/carousel-item.tsx
```diff
@@ -0,0 +1,29 @@
...
+  componentDidLoad() {
+    const parentSwiperContainer = this.host.parentElement.querySelector('swiper-container')
+    const swiperSlide =this.host.querySelector('swiper-slide')
```
felipefialho, line 16: (lint) 
```suggestion
    const swiperSlide = this.host.querySelector('swiper-slide')
```

packages/core/src/components/carousel/carousel-item/carousel-item.tsx
schirrel: @felipefialho i was in doubt about that because we had to add the full doc link to it, and i was afraid of pollute the code

gabrielduete (dismissed):

packages/core/src/components/carousel/carousel-item/carousel-item.tsx
felipefialho: uhmmm maybe we don't need add nothing, just do hahaha

schirrel: sorry @felipefialho i did not get it 😅

schirrel: done [4a70aa3](https://github.com/juntossomosmais/atomium/pull/263/commits/4a70aa341865effe6c181b5d744df087d171bf0b)

felipefialho (approved):

PR #269, opened 2023-09-27
chore: extends eslint from @juntossomosmais/linters

felipefialho:
#### What is being delivered?

- Extends `.eslintrc.js` using `@juntossomosmais/linters/eslintrc.config.js`
- Update `.eslintignore`
- Improve hooks from pre-commit

#### What impacts?

Update `@juntossomosmais/linters` and extends `@juntossomosmais/linters/eslintrc.config.js`

#### Reversal plan

Remove extends


.eslintignore
```diff
@@ -2,5 +2,17 @@ node_modules
...
 dist
 coverage
-!.commitlintrc.js
+.commitlintrc.js
+.eslintrc.js
+.prettierrc.js
 **/*.d.ts
+packages/core/output-target
+packages/core/**/stories
+packages/core/loader
```
tassioFront, line 11: Question: Why ignore those packages too?

felipefialho: `output-target` and `loader` is autogenerated files so is ok

`stories` is a problem because sometimes we need to use specific things to run web components with storybook

What you think?

tassioFront: Sorry, I wasn't clear. 😅  I'm thinking about packages/core/react, packages/core/vue, packages/react and packages/vue

felipefialho: everyone are also auto generated from core, so they cannot pass lint rules

but I'm not sure if we should keep `/stories`, what do you think?

.eslintrc.js
```diff
@@ -0,0 +1,29 @@
...
+        project: './tsconfig.base.json',
+      },
+    },
+  },
+  overrides: [
+    ...baseConfig.overrides,
+    {
+      files: ['**/*.spec.{js,ts,tsx}'],
+      rules: {
+        'sonarjs/no-duplicate-string': 'off',
```
JulioCVaz, line 19: question: make sense add a rule to prevent dependence circle?

like: https://github.com/juntossomosmais/al-khor/blob/master/.eslintrc.js#L15C4-L15C30

package.json
```diff
@@ -19,11 +19,10 @@
     "core:build": "npx nx build @juntossomosmais/atomium && npm run react:build",
     "docs:build": "npx nx build @atomium/docs && npx nx build @atomium/docs-react",
     "docs-react:start": "npx nx run @atomium/docs-react:start",
-    "lint:pre-commit": "eslint --resolve-plugins-relative-to . -c ./.eslintrc.json $(git diff --cached --name-only --diff-filter=AM | grep  -E \"(.js$|.ts$)\") --cache --cache-location node_modules/.cache/eslint",
     "lint:styles": "stylelint '**/*.scss'",
     "icons:build": "npx nx build @juntossomosmais/atomium-icons",
     "postinstall": "npx nx build @juntossomosmais/atomium:prebuild",
-    "pre-commit": "lint-staged",
+    "pre-commit": "git rev-parse -q --no-revs --verify MERGE_HEAD || lint-staged",
```
JulioCVaz, line 25: 🚀

.eslintrc.js
felipefialho: @JulioCVaz Is a good idea add it in our [shared rules](https://github.com/juntossomosmais/time-out-market/blob/main/packages/linters/src/eslint.config.js)? Maybe we could open a PR ❤️

felipefialho: **explaining**: these overrides is just to ignore some rules of `plugin/sonarjs` inside some files

mauriciomutte (dismissed):

igorwessel (approved):
:rocket:

.eslintignore
tassioFront: thanks for the explanation, man! 

> but I'm not sure if we should keep /stories, what do you think?

Yeah... Guess we could test it and then get some feedback.

tassioFront (dismissed):

.eslintignore
felipefialho: Uhmmm, for now I we keep it outside of ignore, if we have some problems we could add again [bc3fab3](https://github.com/juntossomosmais/atomium/pull/269/commits/bc3fab332e843e9cec5baede5098f3bf7cacee96)

tassioFront (dismissed):

tassioFront (dismissed):

igorwessel (approved):

mauriciomutte (dismissed):
🚀

igorwessel (approved):

tassioFront (approved):

PR #290, opened 2023-10-19
feat: create custom render and pre load validations

jowjow22:
## Infos

[Task](https://juntossomosmais.monday.com/boards/1757192123/pulses/5288453411)

#### What is being delivered?

New pagination type being created, the new pagination type is called thumbnails.

#### What impacts?

Affects only the main carousel component when using the paginationType prop as thumbnails.

#### Reversal plan

Revert to the previous stable release.

#### Evidences

image
image


brunosf (approved):

felipefialho:
closes #302

packages/core/src/components/carousel/carousel.scss
```diff
@@ -21,3 +21,42 @@
...
+  display: inline-flex;
+  height: var(--spacing-giant);
+  justify-content: center;
+  opacity: 1;
+  padding: var(--spacing-xxsmall);
+  width: var(--spacing-giant);
+}
+
+::part(thumb-youtube-icon) {
+  color: red;
```
felipefialho, line 44: (block) we need to use variable tokens

packages/core/src/components/carousel/carousel.scss
```diff
@@ -21,3 +21,42 @@
...
+  font-size: 30pt;
```
felipefialho, line 45: (block) why `pt`? we need to use `px` or `em`

packages/core/src/components/carousel/carousel.scss
```diff
@@ -21,3 +21,42 @@
...
+  position: absolute;
```
felipefialho, line 46: (question) this is absolute but what is axis x and y about it? we need to get more control over this element

packages/core/src/utils/renderThumbs.ts
```diff
@@ -0,0 +1,45 @@
+export function renderThumbs(
```
felipefialho, line 1: (block)
- we need to use `kebab-case` to separate file name
- this is a specific thing about the carousel, why is it in global utils?

packages/core/src/utils/renderThumbs.ts
```diff
@@ -0,0 +1,45 @@
...
+  current: number,
+  total: number,
+  urls: Array<string>,
+  videoIcons: boolean
+): string {
+  const thumbs = []
+
+  if (!urls.length) return ''
```
felipefialho, line 9: (question) `url` could be empty?
```suggestion
  if (!urls?.length) return ''
```

felipefialho (changes requested):

packages/core/src/utils/renderThumbs.ts
jowjow22: in this case no, because in the carousel file it has a verify in case that the string is empty, and the parameter is required and well typed, the only case is that someone pass null as param in code

jowjow22: that line always return an array
`const urls = !this.thumbnailImages ? [] : this.thumbnailImages.split(',')`

packages/core/src/utils/renderThumbs.ts
jowjow22: done at [5c47efd](https://github.com/juntossomosmais/atomium/pull/290/commits/5c47efd5dd7f7ec230884f1a402c8fce6453d145)

Removed from global too

packages/core/src/components/carousel/carousel.scss
jowjow22: created the reference in the axis
[5c47efd](https://github.com/juntossomosmais/atomium/pull/290/commits/5c47efd5dd7f7ec230884f1a402c8fce6453d145)

packages/core/src/components/carousel/carousel.scss
jowjow22: i changed to the variable, i forgot to change it since it was one of the first things i did in this feature :sweat: 

[5c47efd](https://github.com/juntossomosmais/atomium/pull/290/commits/5c47efd5dd7f7ec230884f1a402c8fce6453d145)

packages/core/src/components/carousel/carousel.scss
jowjow22: done at [5c47efd](https://github.com/juntossomosmais/atomium/pull/290/commits/5c47efd5dd7f7ec230884f1a402c8fce6453d145)

packages/core/src/utils/renderThumbs.ts
felipefialho: > the only case is that someone pass null as param in code

Don't you think it's safer to maintain optional chaining to avoid any such scenario?

packages/core/src/components/carousel/carousel.tsx
```diff
@@ -41,17 +42,25 @@ export class AtomCarousel {
   @Prop({ mutable: true })
   paginationClickable?: boolean
   @Prop({ mutable: true })
-  paginationType?: PaginationOptions['type'] = 'bullets'
+  paginationType?: PaginationOptions['type'] | 'thumbnails' = 'bullets'
```
brunosf, line 45: (question) this name `thumbnails` is your choice, right? 
In code below is verified and used `custom`, what do you think about using 'custom'

packages/core/src/components/carousel/carousel.tsx
```diff
@@ -64,15 +73,34 @@ export class AtomCarousel {
     this.swiperEl.swiper?.on('navigationNext', () => {
       this.atomClickNext.emit()
     })
+
+    let params: { pagination: PaginationOptions } = this.pagination && {
+      pagination: {
+        type:
+          this.paginationType === 'thumbnails' ? 'custom' : this.paginationType,
```
brunosf, line 80: (suggestion) `custom` instead `thumbnails` not needs this verification

igorwessel (approved):

packages/core/src/components/carousel/carousel.tsx
jowjow22: the custom is because of the default value in the library, so, to not have a generic pagination type in our own carousel, if i use directly custom in the prop, we'll lost this type because it always render the thumbnails, so i believe that makes more sense create a different type that allows use both

packages/core/src/components/carousel/carousel.tsx
jowjow22: like i said in the comment below, i believe that this type is needeed so we dont lose the type custom

Natalie9 (approved):

igorwessel (dismissed):
ótima PR

packages/core/src/components/carousel/carousel.tsx
brunosf: tks for explanation!

brunosf (dismissed):

MayaraRMA (approved):

felipefialho (dismissed):

packages/core/src/components/carousel/stories/carouse.args.ts
```diff
@@ -1,6 +1,5 @@
-import { withActions } from '@storybook/addon-actions/decorator'
```
mauriciomutte, line 1: typo: file name typo `carouse.args.ts` > `carousel.args.ts`

packages/core/src/components/carousel/carousel.tsx
```diff
@@ -64,15 +73,34 @@ export class AtomCarousel {
...
+        clickable: this.paginationClickable,
+      },
+    }
+
+    if (this.paginationType === 'thumbnails') {
+      const urls = !this.thumbnailImages ? [] : this.thumbnailImages.split(',')
```
mauriciomutte, line 86: **suggestion:** you could avoid logical not here
```suggestion
      const urls = this.thumbnailImages ? this.thumbnailImages.split(',') : []
```

packages/core/src/components/carousel/stories/carouse.args.ts
felipefialho: Thanks @mauriciomutte 🙏

packages/core/src/components/carousel/carousel.tsx
jowjow22: done at [0074bf1](https://github.com/juntossomosmais/atomium/pull/290/commits/0074bf15b4fd27067c157150c239c9af6cacd256)

packages/core/src/components/carousel/stories/carouse.args.ts
jowjow22: done at [0074bf1](https://github.com/juntossomosmais/atomium/pull/290/commits/0074bf15b4fd27067c157150c239c9af6cacd256)

mauriciomutte (approved):

felipefialho (approved):

PR #312, opened 2023-11-09
chore: improve buildings using nx

felipefialho:
## Infos

[Task](https://juntossomosmais.monday.com/boards/3957397225/pulses/5479622950)

#### What is being delivered?

- Improve build for all applications using `nx`
- Change CICD to use `nx` to build

#### What impacts?

For now, we start using `dependsOn` to build applications in the right order instead of using a lot of `npm scripts` to do that. 

#### Reversal plan

Fix the wrong build order in `nx.json` and `package.json` files.

### DIFF
--- a/.github/workflows/ghpages.yml
+++ b/.github/workflows/ghpages.yml
@@ -53,12 +53,6 @@ jobs:
       - name: Install dependencies
         run: npm ci --legacy-peer-deps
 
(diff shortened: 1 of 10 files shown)


jowjow22 (approved):

mauriciomutte (approved):
🚀 🚀

PR #360, opened 2023-12-28
fix(core): carousel slides order render

matheus-rosa-jsm:
## Infos

[Task](https://juntossomosmais.monday.com/boards/5148603138/views/113860899/pulses/5752018855)

#### What is being delivered?

The order that the carousel slides were rendered was wrong. This PR fix it passing the responsability of remove slide wrapper to parent.

#### What impacts?

Carousel component takes the responsability to normalize the slides.

#### Reversal plan

#### Evidences

### DIFF
--- a/packages/core/src/components/carousel/carousel.tsx
+++ b/packages/core/src/components/carousel/carousel.tsx
@@ -32,6 +32,8 @@ export class AtomCarousel {
   @Prop({ mutable: true })
   autoplayDelay?: AutoplayOptions['delay']
   @Prop({ mutable: true })
+  centerInsufficientSlides?: boolean = false
+  @Prop({ mutable: true })
   freeMode?: boolean
   @Prop({ mutable: true })
   loop?: boolean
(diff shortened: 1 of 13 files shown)


packages/core/src/components/carousel/carousel.tsx
```diff
@@ -66,6 +66,18 @@ export class AtomCarousel {
 
   componentDidLoad() {
     this.swiperEl = this.host.querySelector('swiper-container')
+
+    const children = this.host.querySelectorAll('atom-carousel-item')
+
+    children.forEach((child) => {
```
felipefialho, line 72: (question) we need to change something in documentation of this component?

matheus-rosa-jsm: Actually No! The component is used exactly in the same way. Just the strategy to normalize atom-carousel-item is diferent

tiago-dela-rosa (approved):

gabrielduete (approved):

PR #399, opened 2024-03-06
fix(button): use default behavior when is link

igorwessel:
## Infos

[Task](https://juntossomosmais.monday.com/boards/1521247929/pulses/6204254489)

#### What is being delivered?

Currently, when we are using the button as a link, it is not possible to have the default behavior because we are preventing it through "onClick".

A small check was added that validates whether it is a link or not.

#### What impacts?

- Atom-Button

#### Reversal plan

- Revert merge

#### Evidences

https://github.com/juntossomosmais/atomium/assets/54173994/84eab80a-3fa1-473e-9f4b-cc7a466e397c

### DIFF
--- a/packages/core/src/components/button/button.spec.ts
+++ b/packages/core/src/components/button/button.spec.ts
@@ -1,4 +1,5 @@
 import { newSpecPage } from '@stencil/core/testing'
+
 import { AtomButton } from './button'
 
 describe('AtomButton', () => {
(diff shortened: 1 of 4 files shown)


packages/core/src/components/button/button.spec.ts
```diff
@@ -173,11 +178,35 @@ describe('AtomButton', () => {
...
+
+    jest.spyOn(event, 'preventDefault')
+    jest.spyOn(event, 'stopPropagation')
+
+    //@ts-expect-error - Testing private method
+    button.handleClick(event)
+
+    expect(event.preventDefault).not.toHaveBeenCalled()
+    expect(event.stopPropagation).not.toHaveBeenCalled()
+  })
```
igorwessel, line 211: I didn't find a way to test this behavior through the spec page, I had to go with the approach suggested by [Stencil](https://stenciljs.com/docs/unit-testing#testing-component-class-logic). 
But we had a small drawback, which was having some warnings related to the mutability of the props. [Issue](https://github.com/ionic-team/stencil/issues/2275)

felipefialho (approved):
Nice fix

gabrielduete (approved):

SallesCosta (approved):

PR #417, opened 2024-04-23
fix(button): adjust button slot align

mauriciomutte:
#### What is being delivered?

- Adjust button slot align to center

#### What impacts?

- Button component with an icon

#### Evidences

| Before | After |
| ----- | ------ | 
| before | after |


igorwessel (approved):

felipefialho (approved):
👏

PR #434, opened 2024-05-10
feat(carousel): add new carousel component

felipefialho:
## Infos

[Task](https://juntossomosmais.monday.com/boards/3957397225/pulses/6587680618)

#### What is being delivered?

- Add a new carousel component
- Move the old carousel as deprecated carousel

#### What impacts?

It PR refactoring completely the carousel component, so it will be necessary to change the way the carousel is used in the project. So we keep for now the old carousel as deprecated carousel, using as `<atom-deprecated-carousel>` to keep the old carousel working.

After refactoring the old carousel inside project (it will be called CarouselShowcase), we will be able to remove the deprecated carousel here.

**advantages:**

- The new carousel will be more flexible and easier to use 🥇 
- Use just vanilla JS, so it will be easier to maintain 🚀 
- The size of the bundle will be reduced as 420kb to 16kb (**96% less**) 😱 

```
 {
    "key": "atom-carousel.entry",
    "components": [
       "atom-carousel"
     ],
     "originalByteSize": 16677
},
{
  "key": "atom-carousel-item.entry",
  "components": [
    "atom-carousel-item"
    ],
  "originalByteSize": 8993
},
{
  "key": "atom-deprecated-carousel.entry",
  "components": [
    "atom-deprecated-carousel"
    ],
    "originalByteSize": 417647
},
{
  "key": "atom-deprecated-carousel-item.entry",
    "components": [
      "atom-deprecated-carousel-item"
    ],
    "originalByteSize": 9022
},
```

#### Evidences

image


packages/core/src/components.d.ts
```diff
@@ -51,24 +51,11 @@ export namespace Components {
...
-        "paginationClickable"?: boolean;
-        "paginationType"?: PaginationOptions['type'] | 'thumbnails';
-        "slidesPerGroup"?: number | string;
-        "slidesPerView"?: number | string;
-        "spaceBetween"?: number;
-        "speed"?: number;
-        "thumbnailImages"?: string;
-        "videoIcons"?: boolean;
-        "watchSlidesProgress"?: boolean;
+        "autoplay": number;
```
tiago-dela-rosa, line 54: (dúvida) Por que precisa de um número? Não pode ser um booleano? Ou é um delay? Na documentação não ficou claro

felipefialho: improve it as [5ec8c32](https://github.com/juntossomosmais/atomium/pull/434/commits/5ec8c321f4c6dca5576e62200cd856ab5d4f2de4)

packages/core/src/components/_deprecated-carousel/carousel.tsx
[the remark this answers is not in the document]
felipefialho: make it sense, thanks for taking this, I'll solve it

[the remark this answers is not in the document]
felipefialho: done at [c5985e8](https://github.com/juntossomosmais/atomium/pull/434/commits/c5985e84a4edbec9b2249e59f3d152db604cc56b)

packages/core/src/components/carousel/stories/carousel.args.ts
```diff
@@ -1,161 +1,78 @@
...
-      table: {
-        category: Category.PROPERTIES,
-      },
-    },
-    thumbnailImages: {
-      description:
-        'String of images URLs to be used as thumbnails splited by comma',
-      defaultValue: { summary: '' },
+      control: 'boolean',
+      description: 'If `true`, the user cannot interact with the button.',
```
MayaraRMA, line 31: I think we can be clearer here. With navigation buttons or pagination buttons?

packages/core/src/components/carousel/stories/carousel.core.stories.tsx
```diff
@@ -1,141 +1,95 @@
...
-  render: () =>
-    createComponent(
-      { pagination: false, slidesPerView: 3, spaceBetween: 40 },
-      'item-default'
-    ),
+export const Loop: StoryObj = {
+  render: (args) => createCarousel(args),
+  args: {
+    ...CarouselComponentArgs,
+    loop: true,
```
MayaraRMA, line 55: Doubt: I didn't understand what this prop does. I didn't see any difference in its behavior

packages/core/src/components/carousel/carousel.tsx
```diff
@@ -1,142 +1,201 @@
...
     }
 
-    if (this.paginationType === 'thumbnails') {
-      const urls = this.thumbnailImages ? this.thumbnailImages.split(',') : []
+    this.currentIndex = newIndex
+    this.showOrHideNavigationButtons()
+  }
+
+  showOrHideNavigationButtons() {
+    if (!this.nextButton || !this.prevButton) return
```
MayaraRMA, line 98: Wouldn't it be the same thing  to use `this.hasNavigation`?

packages/core/src/components/carousel/carousel.tsx
```diff
@@ -1,142 +1,201 @@
...
-      params = {
-        pagination: {
-          ...params.pagination,
-          renderCustom: (_swiper, current, total) =>
-            renderThumbs(current, total, urls, this.videoIcons),
-        },
-      }
+    if (this.loop) {
+      this.nextButton.classList.remove('disabled')
+      this.prevButton.classList.remove('disabled')
```
MayaraRMA, line 102: If `loop` is true, buttons always be visible. Why remove here instead of `componentDidLoad` ? I think this way, you won't need pass for this with every click

packages/core/src/components/carousel/carousel.tsx
```diff
@@ -1,142 +1,201 @@
...
-          this.paginationType === 'thumbnails' ? 'custom' : this.paginationType,
-        clickable: this.paginationClickable,
-      },
+    this.carouselWrapper =
+      this.element.shadowRoot.querySelector('.carousel-wrapper')
+    this.nextButton = this.element.shadowRoot.querySelector('.navigation--next')
+    this.prevButton = this.element.shadowRoot.querySelector('.navigation--prev')
+
+    this.carouselWrapper.addEventListener('touchstart', this.handleTouchStart)
+    this.carouselWrapper.addEventListener('touchend', this.handleTouchEnd)
```
MayaraRMA, line 55: Doubt: What is the difference between `touchstart` and `click`?

felipefialho: `touchstart` works just for touch screen devices, in this case, it is possible change the slide position simulating a swiper because we can check the position of the start and end of the touch

```js
  handleTouchStart = (event: TouchEvent) => {
    this.touchStartX = event.touches[0].clientX
  }

  handleTouchEnd = (event: TouchEvent) => {
    this.touchEndX = event.changedTouches[0].clientX
    this.handleSwipe()
  }

  handleSwipe = () => {
    const lastIndex = this.carouselItems.length - 1
    const isSwipeLeft = this.touchEndX < this.touchStartX

    this.currentIndex = isSwipeLeft
      ? Math.min(this.currentIndex + 1, lastIndex)
      : Math.max(this.currentIndex - 1, 0)

    this.showOrHideNavigationButtons()
  }
```

packages/core/src/components/carousel/stories/carousel.args.ts
felipefialho: Perfect, I improve it here [99b6924](https://github.com/juntossomosmais/atomium/pull/434/commits/99b69242a87e924e5af1311e9b416b827953a876)

packages/core/src/components/carousel/carousel.tsx
felipefialho: No because `!this.nextButton || !this.prevButton` checks the `querySelector` of buttons and `this.hasNavigation` is a prop, but I improved this function here [fe89528](https://github.com/juntossomosmais/atomium/pull/434/commits/fe89528fdb8c8053931afef27f8e9a9c0f78ba07)

packages/core/src/components/carousel/carousel.tsx
felipefialho: Perfect, we don't need to check it, done [fe89528](https://github.com/juntossomosmais/atomium/pull/434/commits/fe89528fdb8c8053931afef27f8e9a9c0f78ba07)

packages/core/src/components/carousel/carousel.tsx
MayaraRMA: if `hasNavegation` is `false`, the `querySelector` will be `undefined`, right?

felipefialho: Yes! But if `hasNavegation` is `true` and happens some error to load `button` `querySelector` will be `undefined` too and create an execution error, so is better to validate all scenarios

packages/core/src/components/carousel/carousel.spec.ts
igorwessel, line 1: nice tests!

packages/core/src/components/carousel/carousel-item/carousel-item.scss
```diff
@@ -1,5 +1,8 @@
 @import '~@atomium/scss-utils/index';
 
-.atom-carousel-item {
-  width: fit-content;
+:host {
+  flex: 1 1 auto;
```
SallesCosta, line 4: 🫡

packages/core/src/components/carousel/carousel.tsx
```diff
@@ -1,142 +1,198 @@
...
+  }
+
+  startAutoplay() {
+    this.loop = true
+
+    setInterval(() => {
+      this.currentIndex = (this.currentIndex + 1) % this.carouselItems.length
+      this.showOrHideNavigationButtons()
+    }, this.autoplay)
+  }
```
igorwessel, line 116: **suggestion:** ‏We can store the loop interval in a variable so we can clear it and prevent memory leaks, in case there are changes in the autoplay. But I imagine this will only happen when people change the autoplay attribute a lot. nevermind I noticed that we only pass it when the component initializes, but we need to clear it when it unmounts.

igorwessel (commented):
Nice! 🚀

packages/core/src/components/carousel/stories/carousel.core.stories.tsx
```diff
@@ -1,141 +1,95 @@
...
+  return html`
+    <atom-carousel
+      loop=${args.loop}
+      autoplay=${args.autoplay}
+      has-navigation=${args.hasNavigation}
+      has-pagination=${args.hasPagination}
+    >
+      <atom-carousel-item>
+        <div
+          style="background-color: var(--color-brand-primary-regular); color: #fff; height: 300px; display: flex; align-items: center; justify-content: center"
```
gabrielduete, line 21: (question): Não conseguimos usar a variável de cor branca no color? `var(--color-neutral-white)` no lugar de `#fff`. Também temos alguns casos no `packages/core/src/components/carousel/stories/carousel.react.stories.tsx`

packages/core/src/components/carousel/carousel.tsx
```diff
@@ -1,142 +1,198 @@
...
+                role='tab'
+                aria-selected={index === this.currentIdx}
+                aria-controls={`carousel-item-${index}`}
+                onClick={() => this.handleMoveToPaginationItem(index)}
+              >
+                {this.thumbnails?.length > 0 ? (
+                  <img
+                    class='thumbnail'
+                    src={this.thumbnails[index]}
+                    alt={`${index + 1}`}
```
gabrielduete, line 186: (question): Não podemos passar somente o valor do `index + 1`?

```suggestion
                    alt={index + 1}
```

packages/core/src/components.d.ts
```diff
@@ -498,27 +521,11 @@ declare namespace LocalJSX {
         "type"?: 'submit' | 'reset' | 'button';
     }
     interface AtomCarousel {
-        "autoplay"?: boolean;
-        "autoplayDelay"?: AutoplayOptions['delay'];
-        "centerInsufficientSlides"?: boolean;
-        "centeredSlides"?: boolean;
-        "freeMode"?: boolean;
+        "autoplay"?: number;
```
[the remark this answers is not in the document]
felipefialho: [same here](https://github.com/juntossomosmais/atomium/pull/434#discussion_r1597064862)

packages/core/src/components/carousel/carousel.tsx
felipefialho: `alt` need to be a string

packages/core/src/components/carousel/stories/carousel.core.stories.tsx
felipefialho: done at [7206eb2](https://github.com/juntossomosmais/atomium/pull/434/commits/7206eb2ed11be3adc839982c341fbc29e045fa9e)

igorwessel (approved):

SallesCosta (approved):

gabrielduete (approved):

MayaraRMA (approved):

PR #461, opened 2024-06-12
feat(list slider): add new component and remove swiper :pray: 

### LINKED ISSUE #445: [Component] List Slider
This slider needs to create item-by-item skipping behavior

felipefialho:
## Infos

[Task](https://juntossomosmais.monday.com/boards/3957397225/pulses/6617725055)

#### What is being delivered?

- Add ListSlider component
- Create `debounce` as utils
- Remove DeprecatedCarrousel

#### What impacts?

It removed completely the DeprecatedCarousel and the lib "Swiper" from the project, so we have a great difference in the final size of the project, it also removed some bugs and dependencies that are not necessary.

*Bundle Sizes* 

```
{
  "key": "atom-deprecated-carousel.entry",
  "components": [
    "atom-deprecated-carousel"
  ],
  "originalByteSize": 417651
},
{
  "key": "atom-list-slider.entry",
  "components": [
     "atom-list-slider"
  ],
  "originalByteSize": 16491
},
```

- **DeprecatedCarousel (using Swiper)**: 417647 bytes 😱 
- **ListSlider**: 16491 bytes 😎 

So if count Carousel (#434) and List Slider together have just 33168 bytes, more than **70% less** than the DeprecatedCarousel

closes #445 

#### Evidences

image


package.json
```diff
@@ -31,8 +31,7 @@
   },
   "dependencies": {
     "@ionic/core": "^7.8.2",
-    "@stencil/core": "^4.17.2",
-    "swiper": "^11.1.1"
```
MayaraRMA, line 35: 🚀

felipefialho: 🙏

brunosf (approved):
:rocket:

anabrook (approved):

MayaraRMA (approved):
🚀

packages/core/src/components/list-slider/list-slider.tsx
```diff
@@ -0,0 +1,260 @@
...
+  }
+
+  handleSwipe = () => {
+    const lastIndex = this.sliderItems.length - 1
+    const isSwipeLeft = this.touchEndX < this.touchStartX
+
+    this.currentIndex = isSwipeLeft
+      ? Math.min(this.currentIndex + 1, lastIndex)
+      : Math.max(this.currentIndex - 1, 0)
+  }
```
tassioFront, line 217: Guess it would be nice to handle all those calculations (mainly in the first load page) in a separate thread. I'm not sure if it is possible, but we can investigate in the future.

tassioFront:
I love that! :rocket:  The slider works so well

igorwessel (approved):
💯

packages/core/src/components/list-slider/list-slider.tsx
felipefialho: Do you have any idea how to improve it? We could try in the future

tassioFront: I had tried Web Workers but directly in the application: https://dev.to/tassiofront/avoid-overloading-the-main-thread-with-web-workers-557c. I'm not sure if it will be able here as Atomium can be used in the server as well... t will be a nice investigation ahahahah

igorwessel: @tassioFront Won't transferring this data between "threads" be worse than calculating directly in the main thread? since it is not very heavy

felipefialho: @igorwessel @tassioFront Could u create an issue to improve it later? 🙏

tassioFront: that's a nice point, @igorwessel ! haha Certainly it needs a balance! 

> @igorwessel @tassioFront Could u create an issue to improve it later?

Sure! I create it and put here :rocket:

PR #489, opened 2024-07-11
feat(icons): add new custom and mdi icons

SallesCosta:
## Infos

[Task](https://juntossomosmais.monday.com/boards/5337789717/views/117729068/pulses/6333243381)

#### What is being delivered?

Describe here all the changes that will be made and what the expected result is.

#### What impacts?

Describe what impacts this delivery has and whether it can cause side effects in other parts of the application.

#### Reversal plan

#### Evidences


packages/core/src/components.d.ts
```diff
@@ -12,738 +12,738 @@ export { IconProps } from "./icons";
 export { Color, Mode, TextFieldTypes } from "@ionic/core";
```
SallesCosta, line 12: Just lint modifications.

packages/icons/svg/mdi/account-check-outline.svg
felipefialho, line 1: question: why the size of icons is completely different?

SallesCosta: @felipefialho 
The svg's source was different. I have changed to keep uniform. [here](https://github.com/juntossomosmais/atomium/pull/489/files)

packages/icons/svg/mdi/account-plus-outline.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="22" height="16" viewBox="0 0 22 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M14 8C16.21 8 18 6.21 18 4C18 1.79 16.21 0 14 0C11.79 0 10 1.79 10 4C10 6.21 11.79 8 14 8ZM14 2C15.1 2 16 2.9 16 4C16 5.1 15.1 6 14 6C12.9 6 12 5.1 12 4C12 2.9 12.9 2 14 2ZM14 10C11.33 10 6 11.34 6 14V16H22V14C22 11.34 16.67 10 14 10ZM8 14C8.22 13.28 11.31 12 14 12C16.7 12 19.8 13.29 20 14H8ZM5 11V8H8V6H5V3H3V6H0V8H3V11H5Z" fill="#7E8B97"/>
```
3rdepaula, line 2: (**block**)We have to remove the attribute fill to use colors directly in the code

3rdepaula (changes requested):

packages/icons/svg/mdi/account-plus-outline.svg
SallesCosta: [removed](https://github.com/juntossomosmais/atomium/pull/489/commits/678f47a20379ed53982993953be08e6873926e37)

3rdepaula (approved):

packages/core/src/components.d.ts
igorwessel: This file is generated automatically, this lint modification will be lost

dyhalmeida (approved):

PR #502, opened 2024-07-29
feat(component): create link component

tassioFront:
## Infos

Creating an atom-link component to be used combined with router components (such as Link, router-link and NuxtLink). Ideally, atom-link takes care of styling as router components work differently. This means we keep needing to import them when creating new anchor links. 

[Task](https://juntossomosmais.monday.com/boards/1757192123/views/36458908/pulses/7117677943)

#### What is being delivered?

Link component and its stories

#### What impacts?

Link components can be used on the projects. Available to test utilizing the version [2.10.0-alpha.2](https://github.com/juntossomosmais/atomium/pkgs/npm/atomium/250778260)

#### Reversal plan

##### React

- Link

[link-Link.webm](https://github.com/user-attachments/assets/1851393f-40fa-48b2-bbed-1b2021d4271f)

- button

[link-btn.webm](https://github.com/user-attachments/assets/8c94d549-4467-46a2-9b18-b0341ae251d3)

##### Vue

- NuxtLink

[link-nuxtlink.webm](https://github.com/user-attachments/assets/24855626-9df6-4f49-af73-416f729b1cf8)

- button

[link-button.webm](https://github.com/user-attachments/assets/785cf4e6-2759-40ba-aa3c-98b6dc607a9a)

### DIFF
--- a/packages/core/loader/cdn.js
+++ b/packages/core/loader/cdn.js
@@ -1,1 +1,3 @@
+
 module.exports = require('../dist/cjs/loader.cjs.js');
+module.exports.applyPolyfills = function() { return Promise.resolve() };
--- a/packages/core/src/components.d.ts
+++ b/packages/core/src/components.d.ts
@@ -143,6 +143,10 @@ export namespace Components {
         "type": TextFieldTypes;
         "value"?: IonTypes.IonInput['value'];
     }
+    interface AtomLink {
+        "color": 'primary' | 'secondary';
+        "type": 'anchor' | 'button';
+    }
     interface AtomListSlider {
         "centralized": boolean;
         "hasNavigation": boolean;
@@ -246,6 +250,10 @@ export interface AtomChipCustomEvent<T> extends CustomEvent<T> {
 export interface AtomInputCustomEvent<T> extends CustomEvent<T> {
     detail: T;
     target: HTMLAtomInputElement;
+}
+export interface AtomLinkCustomEvent<T> extends CustomEvent<T> {
+    detail: T;
+    target: HTMLAtomLinkElement;
 }
 export interface AtomListSliderCustomEvent<T> extends CustomEvent<T> {
     detail: T;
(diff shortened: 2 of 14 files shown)


packages/core/loader/cdn.js
```diff
@@ -1 +1,3 @@
-module.exports = require('../dist/cjs/loader.cjs.js');
\ No newline at end of file
+
```
tassioFront, line 1: It was updated automatically.

packages/core/src/components.d.ts
```diff
@@ -739,6 +771,7 @@ declare module "@stencil/core" {
             "atom-grid": LocalJSX.AtomGrid & JSXBase.HTMLAttributes<HTMLAtomGridElement>;
             "atom-icon": LocalJSX.AtomIcon & JSXBase.HTMLAttributes<HTMLAtomIconElement>;
             "atom-input": LocalJSX.AtomInput & JSXBase.HTMLAttributes<HTMLAtomInputElement>;
+            "atom-link": LocalJSX.AtomLink & JSXBase.HTMLAttributes<HTMLAtomLinkElement>;
```
tassioFront, line 774: It was updated automatically until here.

packages/core/src/components/link/link.tsx
```diff
@@ -0,0 +1,41 @@
...
+    event.stopPropagation()
+
+    return this.click.emit(event)
+  }
+
+  render() {
+    return (
+      <Host>
+        {this.type === 'anchor' ? (
+          <span class='atom-link' color={this.color}>
```
tassioFront, line 25: Not rendering anchor tag as router components (usually) do it - our Next version does not do it, but it is simple to create a wrapper on Colosseum combining Link and anchor tag ([see here](https://juntossomosmais.github.io/atomium/pr-preview/pr-502/?path=/docs/react_components-link--docs))

jowjow22 (approved):

packages/core/src/components/link/link.tsx
felipefialho: maybe we could receive a prop if we can use it as a simple `a` tag?

igorwessel (approved):

packages/core/src/components/link/link.tsx
jowjow22: This is a particularity of each project, considering that the purpose of this DS is to be agnostic.

I think that adding another layer of complexity to this component isn't so necessary, because it gives the choice to the developer and which routing library is he using, leaving to this component only the styling (if it will be used as a literal link anchor) or styling + semantic (if being used as a button who seems like a link)

tassioFront: Yeah, including what Jowjow said, router-link and NuxtLink already render an anchor tag. As it is common for Colosseum combine Link and anchor tag, guess it is simpler to keep atom-link takes care of styling only.

felipefialho (approved):

PR #508, opened 2024-07-31
feat(button): add new fill type to button and link loading

jowjow22:
## Infos

#### What is being delivered?

Adding a new fill type called `outline-filled` to attend styleguide demands.
Add a new property to AtomLink for loading state.

#### What impacts?

AtomButton and AtomLink

#### Reversal plan

Revert to the last version and re-release

#### Evidences

image
image
[Gravação de tela de 31-07-2024 14:25:28.webm](https://github.com/user-attachments/assets/93015c08-9a9c-45bf-8425-fd356e55cfee)


packages/core/src/components/link/link.scss
felipefialho, line 1: block: why are you adding `atom-link` again?

felipefialho (changes requested):
block: @jowjow22 we need to fill the pull request description

jowjow22:
> block: @jowjow22 we need to fill the pull request description

sorry, it should be open as a draft, but maybe I've missed the click

packages/core/src/components/link/link.scss
jowjow22: it isn't being added again, what is being made is a new property to the component to have a loading state, as you asked [here](https://github.com/juntossomosmais/web-customer-typescript/pull/7385#discussion_r1697536051), the version implemented on web-customer is only a way to know that when this version be released, use it on there

packages/core/src/components/button/button.tsx
```diff
@@ -74,9 +74,10 @@ export class AtomButton {
             [`atom-button`]: true,
             [`is-loading`]: this.loading,
             [`is-circle`]: this.shape === 'circle',
+            [`is-outline-filled`]: this.fill === 'outline-filled',
           }}
           color={this.disabled && !this.loading ? 'medium' : this.color}
-          fill={this.fill}
+          fill={this.fill === 'outline-filled' ? 'default' : this.fill}
```
tassioFront, line 80: I didn't understand this validation, man. Why to we need this?

packages/core/src/components/link/link.spec.ts
```diff
@@ -71,4 +71,20 @@ describe('atom-link', () => {
 
     expect(clickEventSpy).toHaveBeenCalled()
   })
+
+  it('should render a clicable button but not emit click event when loading', async () => {
```
tassioFront, line 75: `un`clicable?

packages/core/src/components/button/button.tsx
jowjow22: its for typecheck, the type 'outline-filled' doesnt exists on ion-button, so it throws an error saying that this property cant be bind to this component, doing this, we prevent it

packages/core/src/components/link/link.scss
felipefialho: I think is better add it is another PR, to separate the context

packages/core/src/components/link/link.spec.ts
jowjow22: done at [a9ed8c8](https://github.com/juntossomosmais/atomium/pull/508/commits/a9ed8c83f403c47f1f2c98b2f82db16f09b612b8)

packages/core/src/components/button/button.tsx
tassioFront: hmm got it!

packages/core/src/components/link/link.scss
jowjow22: but even being a small change that doesn't apply any BC? And it wouldn't create 2 releases and 2 new versions?

felipefialho: No, it will be in the same release but in different pull requests. It is important to define the responsibility of each PR for all tasks

packages/core/src/components/button/stories/button.core.stories.tsx
```diff
@@ -44,6 +44,15 @@ export const Secondary: StoryObj = {
...
+export const OutlineFilled: StoryObj = {
+  render: (args) => createButton(args, 'dark'),
+  args: {
+    ...Primary.args,
+    fill: 'outline-filled',
+    color: 'secondary',
+  },
+}
+
 export const White: StoryObj = {
```
tassioFront, line 56: guess we could change this title (White) to `Ghost` or `Outline`

packages/core/src/components/link/link.scss
felipefialho: For this pull request, we could keep it, but is a nice idea separate it in next tasks

packages/core/src/components/button/stories/button.core.stories.tsx
felipefialho: Yes, I think a good idea

jowjow22: done at [69c86f1](https://github.com/juntossomosmais/atomium/pull/508/commits/69c86f1895638bb4b455ccfc85bf7a8f5dc7803b)

tassioFront (approved):

igorwessel (approved):

felipefialho (approved):

PR #563, opened 2024-09-05
feat(modal): add options to disable buttons

matheus-rosa-jsm:
## Infos

[Task](https://juntossomosmais.monday.com/boards/5143488854/views/113762127/pulses/7295837744)

#### What is being delivered?

New optios to disable footer buttons.

#### What impacts?

Describe what impacts this delivery has and whether it can cause side effects in other parts of the application.

#### Reversal plan

#### Evidences

image


felipefialho (dismissed):

igorwessel (approved):

mauriciomutte (dismissed):

tassioFront (dismissed):

felipefialho (dismissed):

packages/core/src/components/modal/modal.tsx
```diff
@@ -24,6 +24,8 @@ export class AtomModal {
   @Prop() alertType?: 'alert' | 'error'
   @Prop() progress?: number
   @Prop() hasFooter?: boolean = true
+  @Prop() disablePrimary?: boolean = false
+  @Prop() disableSecondary?: boolean = false
```
felipefialho, line 28: suggestion: I hadn't noticed but we can remove it

```suggestion
  @Prop() progress?: number
  @Prop() hasFooter = true
  @Prop() disablePrimary = false
  @Prop() disableSecondary = false
```

matheus-rosa-jsm: done

lupy100 (approved):

felipefialho (approved):

brunosf (approved):

PR #572, opened 2024-09-27
fix(modal): add class on mount and fix backdrop

jowjow22:
## Infos

#### What is being delivered?

Fix modal scroll backdrop and expose isOpen Property to control opening of modal without trigger

#### What impacts?

The modal component and it's scroll behaviour

#### Reversal plan

revert to the previous release

#### Evidences

### DIFF
--- a/packages/core/src/components/modal/modal.tsx
+++ b/packages/core/src/components/modal/modal.tsx
@@ -38,7 +38,7 @@ export class AtomModal {
 
   private modal: HTMLAtomModalElement
 
-  private alertMap: AlertType = {
+  private readonly alertMap: AlertType = {
     alert: {
       icon: 'alert-outline',
       color: 'warning',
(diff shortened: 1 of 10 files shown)


packages/core/src/components/modal/modal.tsx
```diff
@@ -58,24 +59,24 @@ export class AtomModal {
...
 
-  private handleSecondaryClick = () => {
+  private readonly handleSecondaryClick = () => {
     this.atomSecondaryClick.emit(this.modal)
   }
 
-  private handlePrimaryClick = () => {
+  private readonly handlePrimaryClick = () => {
     this.atomPrimaryClick.emit(this.modal)
   }
```
jowjow22, line 81: sonar wasnt well configured, so now that is checking code on the right way, it throws an error of the functions and properties never being reassigned, so they can be marked as readonly

tassioFront (approved):

lupy100 (approved):

tiago-dela-rosa (approved):

felipefialho (approved):

PR #590, opened 2024-10-09
feat(select): add validation in select tag

Natalie9:
## Infos

[Task](https://juntossomosmais.monday.com/boards/3345740155/views/76424285/pulses/7170873692)

#### What is being delivered?

Adds a validation when the element to attach the tag to is not found in the select options.

#### What impacts?

Select component

#### Reversal plan

Rollback this PR

#### Evidences

Sentry triggered some errors related to this code snippet and with potential navigation crash. These errors were not reproducible because they occur on specific devices. However, to work around the problem and prevent navigation crashes, the changes in this PR were made.

There were no visual changes. The change occurs if there is any issue when searching for the element to attach the tag and it is not found, return to the view without adding the tag.


packages/core/src/components/select/select.tsx
```diff
@@ -71,8 +71,14 @@ export class AtomSelect {
       const optionElement =
         this.getElementByTag(itemElement, 'ion-radio') ||
         this.getElementByTag(itemElement, 'ion-checkbox')
+
+      if (!optionElement?.shadowRoot) return
```
tassioFront, line 75: were you able to reproduce it?

Natalie9: I couldn't reproduce the error, so I decided to make this change and go with the production testing approach. LOL.

tassioFront: That's ok. Did you create another task to figure out what's going on? then we can merge this first PR and explore the root case into another.

tassioFront (approved):

gabrielduete (approved):

igorwessel (approved):

mauriciomutte (approved):

packages/core/src/components/select/select.tsx
Natalie9: created here https://juntossomosmais.monday.com/boards/3345740155/views/76424285/pulses/7675362982

PR #611, opened 2024-11-06
fix(select): add async to method to avoid promise warning

felipefialho:
#### What is being delivered?

- Add async to method to add tag to select 
- Fix docs to open select with tags

#### Evidences

image

### DIFF
--- a/packages/core/src/components/select/select.tsx
+++ b/packages/core/src/components/select/select.tsx
@@ -5,10 +5,10 @@ import {
   Element,
   Event,
   EventEmitter,
-  Host,
-  Prop,
   h,
+  Host,
   Method,
+  Prop,
 } from '@stencil/core'
 
 import { IconProps } from '../../icons'
@@ -50,7 +50,7 @@ export class AtomSelect {
   @Event() atomFocus!: EventEmitter<void>
 
   @Method()
-  setTagInSelectOptions() {
+  async setTagInSelectOptions() {
     /**
      * This method was necessary because the `ion-selection-option` loop does not allow customizations or custom components.
      * So, to be able to add custom elements such as a tag or a badge inside an option of the `select` field, when the select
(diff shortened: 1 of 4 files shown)


packages/core/src/components/select/select.tsx
```diff
@@ -50,7 +50,7 @@ export class AtomSelect {
   @Event() atomFocus!: EventEmitter<void>
 
   @Method()
-  setTagInSelectOptions() {
+  async setTagInSelectOptions() {
```
tassioFront, line 53: do you know why it is complaining? I was seeing the code and I don't see any promise :thinking:

igorwessel: @tassioFront stenci requires all methods to be asynchronous because its entire API is asynchronous. Even if there is no promise, we need to make them asynchronous. 😆 
I discovered this today while working on the tooltip. :haha:

igorwessel (approved):

packages/core/src/components/select/select.tsx
felipefialho: Exactly @tassioFront, it was causing a warning during the build 😆

packages/core/src/components/select/stories/select.core.stories.tsx
```diff
@@ -104,13 +104,18 @@ export const Multiple: StoryObj = {
...
-  ...optionsDefault,
   {
-    id: '3',
+    id: '1',
     value: 'Nice Green',
     disabled: false,
     tag: { color: 'success', label: 'New ' },
   },
+  {
+    id: '2',
```
tassioFront, line 114: :raised_hands:

packages/core/src/components/select/select.tsx
tassioFront: dammn, I didn't know that haha thanks!

tassioFront (approved):

PR #617, opened 2024-11-11
fix(modal): change variable's name on modal to zindex

jowjow22:
## Infos

#### What is being delivered?

Setting the new value for z-index on atom modal, the motivation is that ionic sets it too high and most of other UI libraries doesn't use this value (the value for it before was 20002) 

#### What impacts?

Atom modal component styling and Atom steps modal component

#### Reversal plan

Re release of the previous version and revert on main branch

#### Evidences

### DIFF
--- a/packages/core/src/components/modal/modal.scss
+++ b/packages/core/src/components/modal/modal.scss
@@ -1,7 +1,12 @@
 @import '~@atomium/scss-utils/index';
 
+:host {
+  --atom-modal-z-index: var(--zindex-1000);
+}
+
 .atom-modal {
   position: fixed;
+  z-index: var(--atom-modal-z-index) !important;
 
   &__close {
     position: absolute;
--- a/packages/core/src/components/modal/modal.scss
+++ b/packages/core/src/components/modal/modal.scss
@@ -1,12 +1,12 @@
 @import '~@atomium/scss-utils/index';
 
 :host {
-  --atom-modal-z-index: var(--zindex-1000);
+  --atom-modal-zindex: var(--zindex-1000);
 }
 
 .atom-modal {
   position: fixed;
-  z-index: var(--atom-modal-z-index) !important;
+  z-index: var(--atom-modal-zindex) !important;
 
   &__close {
     position: absolute;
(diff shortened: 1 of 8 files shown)


packages/core/src/components/modal/modal.scss
```diff
@@ -1,12 +1,12 @@
 @import '~@atomium/scss-utils/index';
 
 :host {
-  --atom-modal-z-index: var(--zindex-1000);
+  --atom-modal-zindex: var(--zindex-1000);
```
felipefialho, line 4: we need to update docs

jowjow22: even if it's a local variable? i've followed the atom-tag example and it didn't had a example on docs :thinking:

felipefialho: [like a select example](https://juntossomosmais.github.io/atomium/?path=/docs/components-select--docs)

image

jowjow22: done it here!! [5efe8ab](https://github.com/juntossomosmais/atomium/pull/617/commits/5efe8abba05d32cdea8b157d550e546ffdf78bb1)

jowjow22: thanks for the example

packages/core/src/components/modal/stories/modal.args.ts
```diff
@@ -160,6 +160,13 @@ export const ModalStoryArgs = {
         category: Category.SLOTS,
       },
     },
+    '--atom-modal-zindex': {
```
felipefialho, line 163: we need to add it to docs of `atom-steps-modal` too

jowjow22: done at [ff7cc0d](https://github.com/juntossomosmais/atomium/pull/617/commits/ff7cc0d21f35c136c5f8de648288ea63ff2e289a)

felipefialho (approved):

tassioFront (approved):

PR #627, opened 2024-11-21
feat(atom-steps-modal): initial step feature

jowjow22:
## Infos

#### What is being delivered?

adding new property to allow user to start the modal directly in a specific step even though it has more steps behind.
Change the use of primaryButtonText and secondaryButtonText to be an array of string for each step, like the titles.
Change the name of this properties to be more descriptive with the purpose

#### What impacts?

BREAKING CHANGE: Atom steps modal component and primaryButton and secondaryButton texts functionalities, the primaryButtonText and secondaryButtonText properties don't will exists by now on.

#### Reversal plan

Re-release of the previous version and revert on main branch.

#### Evidences

[Gravação de tela de 22-11-2024 11:21:28.webm](https://github.com/user-attachments/assets/1a91fce3-5811-4981-8636-aaeb431dec82)

### DIFF
--- a/packages/core/src/components.d.ts
+++ b/packages/core/src/components.d.ts
@@ -195,13 +195,15 @@ export namespace Components {
         "value"?: IonTypes.IonSelect['value'];
     }
     interface AtomStepsModal {
+        "cancelButtonText"?: string;
         "closeOnFinish"?: boolean;
         "currentStep": number;
+        "customInitialStep"?: number;
         "disablePrimaryButton"?: boolean;
         "disableSecondaryButton"?: boolean;
         "isOpen": boolean;
-        "primaryButtonText"?: string;
-        "secondaryButtonText"?: string;
+        "primaryButtonTextsByStep": string;
+        "secondaryButtonTextsByStep": string;
         "steps": number;
         "stepsTitles": string;
         "trigger"?: string;
@@ -825,8 +827,10 @@ declare namespace LocalJSX {
         "value"?: IonTypes.IonSelect['value'];
     }
     interface AtomStepsModal {
+        "cancelButtonText"?: string;
         "closeOnFinish"?: boolean;
         "currentStep"?: number;
+        "customInitialStep"?: number;
         "disablePrimaryButton"?: boolean;
         "disableSecondaryButton"?: boolean;
         "isOpen"?: boolean;
@@ -838,8 +842,8 @@ declare namespace LocalJSX {
         "onAtomIsOpenChange"?: (event: AtomStepsModalCustomEvent<any>) => void;
         "onAtomNextStep"?: (event: AtomStepsModalCustomEvent<any>) => void;
         "onAtomPreviousStep"?: (event: AtomStepsModalCustomEvent<any>) => void;
-        "primaryButtonText"?: string;
-        "secondaryButtonText"?: string;
+        "primaryButtonTextsByStep"?: string;
+        "secondaryButtonTextsByStep"?: string;
         "steps"?: number;
         "stepsTitles"?: string;
         "trigger"?: string;
--- a/packages/core/src/components/steps-modal/steps-modal.spec.ts
+++ b/packages/core/src/components/steps-modal/steps-modal.spec.ts
@@ -348,7 +348,7 @@ describe('atom-steps-modal', () => {
             steps="3"
             trigger="open-modal-steps"
             steps-titles="Step 1, Step 2, Step 3"
-            custom-initial-step="2"
+            locked-initial-step="2"
             primary-button-texts-by-step="Next, Next, Finish"
             secondary-button-texts-by-step="Close, Close, Previous"
         >
@@ -369,7 +369,7 @@ describe('atom-steps-modal', () => {
     expect(cancelSpy).toHaveBeenCalled()
   })
 
-  it('should adjust progress when customInitialStep is set', async () => {
+  it('should adjust progress when locked-initial-step is set', async () => {
     page = await newSpecPage({
       components: [AtomStepsModal],
       html: `
@@ -378,7 +378,7 @@ describe('atom-steps-modal', () => {
             steps="3"
             trigger="open-modal-steps"
             steps-titles="Step 1, Step 2, Step 3"
-            custom-initial-step="2"
+            locked-initial-step="2"
             primary-button-texts-by-step="Next, Next, Finish"
             secondary-button-texts-by-step="Close, Close, Previous"
         >
@@ -390,8 +390,8 @@ describe('atom-steps-modal', () => {
 
     expect(page.rootInstance.progress).toBe(0.5)
   })
-  
-  it('should set customInitialStep to current step when dismiss is called', async () => {
+
+  it('should set locked-initial-step to current step when dismiss is called', async () => {
     page = await newSpecPage({
       components: [AtomStepsModal],
       html: `
@@ -400,7 +400,7 @@ describe('atom-steps-modal', () => {
             steps="3"
             trigger="open-modal-steps"
             steps-titles="Step 1, Step 2, Step 3"
-            custom-initial-step="2"
+            locked-initial-step="2"
             primary-button-texts-by-step="Next, Next, Finish"
             secondary-button-texts-by-step="Close, Close, Previous"
         >
--- a/packages/core/src/components/steps-modal/steps-modal.spec.ts
+++ b/packages/core/src/components/steps-modal/steps-modal.spec.ts
@@ -286,7 +286,7 @@ describe('atom-steps-modal', () => {
       page.root?.querySelector('atom-modal')?.getAttribute('header-title')
     ).toBe('Step 2')
   })
-  
+
   it('should close modal when closeOnFinish is set', async () => {
     page = await newSpecPage({
       components: [AtomStepsModal],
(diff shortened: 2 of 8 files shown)


anabrook (dismissed):

luiz-ortega (dismissed):

packages/core/src/components.d.ts
```diff
@@ -825,8 +827,10 @@ declare namespace LocalJSX {
         "value"?: IonTypes.IonSelect['value'];
     }
     interface AtomStepsModal {
```
felipefialho, line 829: question: I know that this file is generated automatically but is weird `interface AtomStepsModal` generating twice, do you know what?

jowjow22: i think that is the way stencil create those interfaces, because if u see, all the component interfaces are created twice, one with the types being in the way on the component, other being all props setted as optional

packages/core/src/components/steps-modal/steps-modal.tsx
```diff
@@ -85,6 +95,13 @@ export class AtomStepsModal {
   }
 
   private handleSecondaryClick = () => {
+    if (this.currentStep === this.customInitialStep) {
+      this.isOpen = false
+      this.atomCancel.emit()
+
```
tassioFront, line 101: Guess it is not a steps-modal responsibility :thinking:

packages/core/src/components/steps-modal/steps-modal.tsx
```diff
@@ -105,7 +122,25 @@ export class AtomStepsModal {
     e.stopImmediatePropagation()
     this.isOpen = false
     this.atomDidDismiss.emit(this.currentStep)
-    this.currentStep = 1
+    this.currentStep = this.customInitialStep ? this.customInitialStep : 1
```
tassioFront, line 125: suggestion:

` this.currentStep = this.customInitialStep ?? 1`

packages/core/src/components/steps-modal/steps-modal.tsx
```diff
@@ -105,7 +122,25 @@ export class AtomStepsModal {
...
+  }
+
+  private get secondaryButtonText() {
+    return this.currentStep === this.customInitialStep
```
tassioFront, line 129: Shouldn't we remove the `cancelButtonText` as there is the `secondaryButtonTextsArray`?

packages/core/src/components/steps-modal/steps-modal.tsx
```diff
@@ -105,7 +122,25 @@ export class AtomStepsModal {
...
+      : this.secondaryButtonTextsArray[this.currentStep - 1].trim()
+  }
+
+  private get primaryButtonText() {
+    return this.primaryButtonTextsArray[this.currentStep - 1].trim()
+  }
+
+  private get progress() {
+    if(this.customInitialStep) {
+      return (this.currentStep - this.customInitialStep + 1) / (this.steps + 1 - this.customInitialStep)
```
tassioFront, line 140: we need some constants to explain this math, man

packages/core/src/components/steps-modal/steps-modal.tsx
jowjow22: the cancel button text is a facilitator to allow user to set the secondary button text directly when it is on first step, so the user doesn't need to set a new array of strings for each time the customInitialStep is set, they only set the default cases and this cancel, so they can change the custom initial step to wherever he wants without need a new array

packages/core/src/components/steps-modal/steps-modal.tsx
jowjow22: done at[f814758](https://github.com/juntossomosmais/atomium/pull/627/commits/f81475800acd02c38b0999983bb92aff8b013f53)

packages/core/src/components/steps-modal/steps-modal.tsx
jowjow22: done at [f814758](https://github.com/juntossomosmais/atomium/pull/627/commits/f81475800acd02c38b0999983bb92aff8b013f53)

packages/core/src/components/steps-modal/steps-modal.tsx
jowjow22: yeah, i forgot that  when is on first step and clicks secondary It should only emit cancel

packages/core/src/components/steps-modal/steps-modal.tsx
jowjow22: fixed the calc here [d8892bc](https://github.com/juntossomosmais/atomium/pull/627/commits/d8892bc7e0c1a5ebd1b5e3e2c7a05c613fa4c5fd)

tassioFront: stepsQuantityA`d`justedToCustomInitial is missing the `d`

packages/core/src/components/steps-modal/steps-modal.tsx
tassioFront: Sorry, man. I don't know if I got your point. Once the array is given, isn't always passed anyway?

packages/core/src/components/steps-modal/steps-modal.tsx
```diff
@@ -105,7 +121,28 @@ export class AtomStepsModal {
...
+  private get primaryButtonText() {
+    return this.primaryButtonTextsArray[this.currentStep - 1].trim()
+  }
+
+  private get progress() {
+    if(this.customInitialStep) {
+      const currentStepAdjustedAsInitial = this.currentStep - this.customInitialStep + 1
+      const stepsQuantityAjustedToCustomInitial = (this.steps + 1 - this.customInitialStep)
+
+      return ( currentStepAdjustedAsInitial / stepsQuantityAjustedToCustomInitial )
```
felipefialho, line 142: lint: didn't work automatically?

we also could use `early return`

```suggestion
    if (!this.customInitialStep)
    
    const currentStepAdjustedAsInitial = this.currentStep - this.customInitialStep + 1
    const stepsQuantityAjustedToCustomInitial = (this.steps + 1 - this.customInitialStep)

    return (currentStepAdjustedAsInitial / stepsQuantityAjustedToCustomInitial)
```

packages/core/src/components.d.ts
igorwessel: i do not know the why, but i think it's duplicated interface because namespace. It must create a representation of the AtomModal in a specific namespace, type: global, HTML, etc...

packages/core/src/components/steps-modal/steps-modal.tsx
igorwessel: To make the calculation more simple, doesn't it make sense to receive the `customInitialStep` without having to pass the number as an index? Since the entire calculation is not based on the index.

igorwessel: maybe we don't need a diferrent calculation, Isn't currentStep already set to use customInitialStep?

packages/core/src/components/steps-modal/steps-modal.spec.ts
```diff
@@ -332,4 +338,86 @@ describe('atom-steps-modal', () => {
     expect(isOpenChangeSpy).toHaveBeenCalledTimes(2)
     expect(isOpenChangeSpy.mock.calls[1][0].detail).toBe(false)
   })
+
+  it('should emit atom cancel the modal when customInitialStep is set and the secondary button is clicked on that step', async () => {
```
igorwessel, line 342: praise: nice

packages/core/src/components/steps-modal/steps-modal.tsx
jowjow22: the `customInitialStep` feature was made to ignore the previous steps existing before the index of this prop, so if I create a steps modal with 4 steps, but set `customInitialStep` as 3, the first step showing when opening the modal should be 3 and, as now, the modal, visually only have 2 steps, the progress should change, instead of being a 4 step progress (1/4+2/4+3/4+4/4) it should be only a 2 step progress (1/2+2/2), so to maintain the coherence between this change, this calculation is needed


And about what @felipefialho about early return, i didn't see any diff between this two approaches, since we need a return of the adjusted progress and a return of the default progress, currently I'm already using early return

igorwessel: >the customInitialStep feature was made to ignore the previous steps existing before the index of this prop, so if I create a steps modal with 4 steps, but set customInitialStep as 3, the first step showing when opening the modal should be 3 and, as now, the modal, visually only have 2 steps, the progress should change, instead of being a 4 step progress (1/4+2/4+3/4+4/4) it should be only a 2 step progress (1/2+2/2), so to maintain the coherence between this change, this calculation is needed

@jowjow22  But anyway, it’s still a 4-step modal, right? Or is the idea actually to cut the previous steps?

jowjow22: @felipefialho did the early return here [a44c81d](https://github.com/juntossomosmais/atomium/pull/627/commits/a44c81d727fc3605a6e972dfcaa8d78cc2fdffaa)

igorwessel: @jowjow22 Oh, maybe we could change the name for: `startModalOnStep`. Currently, I have the impression that will be the current step.

igorwessel (approved):

tassioFront (approved):

PR #650, opened 2024-12-06
fix(pagination): edge case when count is equal a 1

igorwessel:
## Infos

#### What is being delivered?

Minor fix on pagination when page count is less or equal a one, the bug in question was due to boundary pages.

#### What impacts?

- Pagination

#### Reversal plan

- Revert PR

#### Evidences

Captura de Tela 2024-12-06 às 13 26 25

Captura de Tela 2024-12-06 às 13 27 01

### DIFF
--- a/packages/core/src/components/pagination/pagination.tsx
+++ b/packages/core/src/components/pagination/pagination.tsx
@@ -38,6 +38,12 @@ export class AtomPagination {
   makeItems() {
     const items = []
 
+    if (this.count <= 1) {
+      items.push({ type: 'page', number: 1 })
+
+      return items
+    }
+
     const siblingsStart = Math.max(
       Math.min(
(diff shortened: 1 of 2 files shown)


packages/core/src/components/pagination/pagination.tsx
```diff
@@ -38,6 +38,12 @@ export class AtomPagination {
   makeItems() {
     const items = []
 
+    if (this.count <= 1) {
+      items.push({ type: 'page', number: 1 })
+
+      return items
+    }
```
igorwessel, line 45: **thought:** ‏I could move this solution into the calc but I thought it would only complicate things, since there is a very simple solution.

tassioFront (approved):

dyhalmeida (approved):

Issue #535, opened 2024-08-13
[Fix] Modal workaround


matheus-rosa-jsm:
## The problem

In the version 7.0 of ionic ion-modal has an undesired behavion on backdrop. It block body scrolling.
Inonic issue: https://github.com/ionic-team/ionic-framework/issues/23942
This issue was resolved in the new ionic maijor version, 8.0

## What is needed
To avoid this behavior was created an workaround and it will be no longer needed after the ionic upgrade. So is necessary remove it.

PR with the workaround: https://github.com/juntossomosmais/atomium/pull/534

It's necessary remove all manual interaction with backdrop-no-scroll class:
`document.body.classList.remove(BACKDROP_NO_SCROLL)`

tassioFront:
We needed to improve the workaround as more issues were happening. [The last adjustments are found here](https://github.com/juntossomosmais/atomium/pull/572).

tassioFront:
@matheus-rosa-jsm @jowjow22 , we can review this issue once ionic 8.0 is published on atomium. Let's talk! 

[Linking](https://github.com/juntossomosmais/atomium/pull/684#discussion_r1907688927)

FYI @felipefialho

matheus-rosa-jsm:
Good! I'm working on it @tassioFront.

PR #684, opened 2025-01-08
fix(core): increase priority of custom ionic css variables

felipefialho:
#### What is being delivered?

- Increase priority of custom Ionic CSS variables
- Add `shadow: true` for modal
- Add `scss-utils` implicit for all `scss` files

#### What impacts?

Components like Modal in Ionic 8 changed the way they are loaded, for now, they are added to the DOM on demand.

As on Atomium, we are using an individual import of `@ionic/core` by each component. When a new component is added to the DOM, it overwrites the custom CSS variables. To solve it we need to increase the priority of custom CSS variables using `body` instead of `:root`.


anabrook (approved):

anabrook (dismissed):

packages/core/src/components/modal/modal.tsx
```diff
@@ -68,11 +67,6 @@ export class AtomModal {
   }
 
   componentDidLoad() {
-    /* @todo it's needed to prevent a ionic error. In the version 8.0 it was fixed, remove it after the upgrade.
```
tassioFront, line 71: BACKDROP_NO_SCROLL is used in more places on this file, guess it can be done after migration in a dedicated task. What do you think?

packages/core/src/components/steps-modal/steps-modal.tsx
```diff
@@ -12,7 +12,6 @@ import {
 @Component({
   tag: 'atom-steps-modal',
   styleUrl: 'steps-modal.scss',
-  shadow: false,
```
tassioFront, line 15: :rocket:

packages/core/src/components/modal/modal.tsx
felipefialho: perfect! I [return it]([7ab6f48](https://github.com/juntossomosmais/atomium/pull/684/commits/7ab6f487bbb98d2ca5592eaa7d5f4f8d7b9edc8e)). could u check it with @jowjow22 and @matheus-rosa-jsm next week?

packages/core/stencil.config.ts
```diff
@@ -114,7 +114,10 @@ export const config: Config = {
   namespace: 'core',
   plugins: [
     sass({
-      injectGlobalPaths: ['../../node_modules/@ionic/core/css/core.css'],
+      injectGlobalPaths: [
```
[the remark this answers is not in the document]
igorwessel: :rocket:

igorwessel (approved):

anabrook (approved):

packages/core/src/components/modal/modal.tsx
tassioFront: I just [commented](https://github.com/juntossomosmais/atomium/issues/535#issuecomment-2578480361)  on the issue to remember us. ;)

tassioFront (approved):

PR #687, opened 2025-01-10
feat(datetime): add new component

felipefialho:
#### What is being delivered?

- Add `atom-datetime` component
- Add `atom-datetime` tests
- Add `atom-datetime` from storybook

#### What impacts?

A new component `Datetime` that will be used to display date and time in a consistent way across the application, this is a wrapper of `ion-datetime` [component from Ionic](https://ionicframework.com/docs/api/datetime#properties)

We also add a prop called `rangeMode` that will allow the user to select a range of dates.

image


MayaraRMA (dismissed):

packages/core/src/components/datetime/datetime.scss
```diff
@@ -0,0 +1,102 @@
...
+  transform: translateY(-50%);
+  z-index: var(--zindex-5);
+}
+
+.atom-button {
+  border-bottom: 1px solid
+    var(--ion-color-step-500, var(--ion-background-color-step-500, gray));
+  width: 100%;
+
+  .atom-item:hover & {
```
tassioFront, line 58: is there an extra `&`? :thinking:

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -0,0 +1,244 @@
...
+      day: '2-digit',
+      year: 'numeric',
+    },
+    time: {
+      hour: '2-digit',
+      minute: '2-digit',
+    },
+  }
+  @Prop() highlightedDates?: DatetimeHighlight[] | DatetimeHighlightCallback
+  @Prop() hourCycle: 'h12' | 'h23' = 'h23'
```
tassioFront, line 50: should it be 'h24'?

anabrook (dismissed):

packages/core/src/components/datetime/datetime.scss
felipefialho: Nops! The propose here is catch the `atom-button` when it is inside `.atom-item:hover`, this is the generated CSS

```css
.atom-item:hover .atom-button::part(native) {
  background: var(--ion-color-step-100, var(--ion-background-color-step-100, #e6e6e6));
}
```

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -0,0 +1,244 @@
...
+
+  private readonly handleFocus = () => {
+    this.atomFocus.emit()
+  }
+
+  private readonly handleCancel = () => {
+    this.atomCancel.emit()
+  }
+
+  private calculateDateRange(dates: string[]): string[] {
```
tassioFront, line 111: I didn't understand why it needed this pre-math, to be honest :thinking:  I mean, the date conversion does, but I didn't get the incremental part.

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -0,0 +1,244 @@
...
+          ? dates
+          : this.selectedDates.concat(dates)
+    } else {
+      this.selectedDates = []
+    }
+
+    this.atomChange.emit(this.selectedDates)
+  }
+
+  private handleDateChange = (event) => {
```
tassioFront, line 143: I found this for typing, man :heart_eyes: 

```TS
import { JSX } from '@ionic/core/dist/types/components'

JSX.IonDatetime['onIonChange']
```

packages/core/src/components/datetime/datetime.scss
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  &.date-active {
+    border-bottom-color: var(--color-brand-secondary-regular);
+
+    & ~ .atom-label {
+      color: var(--color-brand-secondary-regular);
+    }
+
+    & ~ .atom-icon {
```
tassioFront, line 98: nice use of ~ :rocket: :rocket: it helps a lot

packages/core/src/components/datetime/datetime.scss
tassioFront: daammn, I didn't know that. Thanks!

packages/core/src/components/datetime/datetime.tsx
tassioFront: what the crazy, dude. Thanks!

packages/core/src/components/datetime/datetime.tsx
felipefialho: This pre-math and increment ensure the function works for both increasing and decreasing date ranges. The increment determines direction, `daysDifference` sets the correct range length, and `setDate` adjusts each step. This guarantees all dates are included, regardless of order

This is the most secure way to avoid errors with some data intervals

packages/core/src/components/datetime/datetime.tsx
felipefialho: Wow, I didn't know about it! Thanks a lot

[2cb1ff5](https://github.com/juntossomosmais/atomium/pull/687/commits/2cb1ff581c27a4532e8d2a0ab8f84178fa01f23a)

packages/core/src/components/datetime/datetime.tsx
tassioFront: Got it. I was surprised to see math like that, it looks like something expected from the own component.

felipefialho: Ahhh I understand your point now, the problem is because `ion-datetime` [supports multiple date selection] but doesn't have a range of dates 😢 , then we would create this calc every time that we need to use it, to avoid adding more complexity inside our projects I created this abstraction here

tassioFront: hmmm I see this now, you told us. Yeah, man, better to keep it here and avoid bugs on the other apps!

MayaraRMA (approved):

PR #726, opened 2025-03-21
feat(component): create stepper component

gabrielduete:
[Task](https://juntossomosmais.monday.com/boards/3345740155/views/76424285/pulses/9025163098)

Close: #108 

## Summary

This pull request introduces a new `AtomStepper` component, a visual representation of user progress through a series of steps. It includes the component's implementation, styling, unit tests, and Storybook stories for React, Vue, and Web Components. Below are the most important changes grouped by theme:

### Component Implementation:

* Added the `AtomStepper` component in `stepper.tsx`, supporting properties like `steps`, `activeStep`, `stepTitle`, and `disabledUnrenderedStep`. It dynamically renders steps with different states (active, completed, disabled) and includes a mobile-friendly layout.
* Introduced the `isMaxTablet` utility function in `screens.ts` to detect tablet-sized screens for responsive design.

### Styling:

* Created a new SCSS file, `stepper.scss`, defining styles for the stepper, including states such as active, completed, and disabled, as well as mobile-specific styles.

### Testing:

* Added unit tests in `stepper.spec.ts` to validate the rendering of steps, including active, completed, and disabled states, as well as the mobile view.

### Documentation and Storybook:

* Added Storybook stories for the `AtomStepper` component in React (`stepper.react.stories.tsx`), Vue (`stepper.vue.stories.tsx`), and Web Components (`stepper.core.stories.tsx`), showcasing different states like in-progress, uncompleted, and completed. [[1]](diffhunk://#diff-c4be458f1506852496ac89b4205847236252d90f32bf0c6935c964531d0aba7dR1-R58) [[2]](diffhunk://#diff-58b6afbe1c6d8907059d62eff6f63eb694e9d3d65f72227787d22781bee7b893R1-R61) [[3]](diffhunk://#diff-0b9180e2981b3dfa84c86ba8e86c6ac1b448b409f051d026655f2f79214783b4R1-R57)
* Defined reusable arguments and documentation for the component in `stepper.args.ts`, including usage guidelines and property descriptions.

## Evidences

Desktop

https://github.com/user-attachments/assets/f2e9ab09-6dbc-4555-aa89-2222aa2c0b32

Mobile

image


brunosf (dismissed):

packages/core/src/components/stepper/stepper.tsx
```diff
@@ -0,0 +1,113 @@
...
+    this.parsedSteps = typeof newVal === 'string' ? JSON.parse(newVal) : newVal
+  }
+
+  componentWillLoad() {
+    if (typeof this.steps === 'string') {
+      this.parsedSteps = JSON.parse(this.steps)
+    } else {
+      this.parsedSteps = this.steps
+    }
+  }
```
MayaraRMA, line 38: I didn't understand where do this steps come from?

packages/core/src/components/stepper/stepper.tsx
```diff
@@ -0,0 +1,113 @@
...
+  @Prop() disabledUnrenderedStep?: boolean = true
+
+  private parsedSteps: {
+    title: string
+    completed: boolean
+  }[] = []
+
+  @Watch('steps')
+  parseSteps(newVal) {
+    this.parsedSteps = typeof newVal === 'string' ? JSON.parse(newVal) : newVal
```
MayaraRMA, line 29: Why do you have to do this?

packages/core/src/components/stepper/stepper.scss
```diff
@@ -0,0 +1,118 @@
...
+  }
+}
+
+.title {
+  color: var(--color-neutral-regular);
+  font: var(--text-body-small);
+  letter-spacing: var(--text-body-small-letter);
+}
+
+.active {
```
tassioFront, line 71: We have used the is/has pattern for these cases. Guess it should be be:
 
```CSS
.title {

&.is-active {
// do something
}
}
```

packages/core/src/components/stepper/stories/stepper.args.ts
```diff
@@ -0,0 +1,49 @@
+export const StepperStoryArgs = {
+  decorators: [],
+  parameters: {
+    actions: {
+      handles: [],
+    },
+    docs: {
+      description: {
+        component: `The Stepper is a visual representation of the user's progress through a set of steps, guiding them toward completion of a flow.
```
tassioFront, line 9: good suggestions!

packages/core/src/components/stepper/stepper.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  componentWillLoad() {
+    if (typeof this.steps === 'string') {
+      this.parsedSteps = JSON.parse(this.steps)
+    } else {
+      this.parsedSteps = this.steps
+    }
+  }
+
+  private readonly isCompleted = (completed: boolean) =>
```
tassioFront, line 40: As it doesn't return a boolean, you can change to something like: addCompletedClass/shouldAddCompletedClass etc

packages/core/src/components/stepper/stepper.tsx
```diff
@@ -0,0 +1,113 @@
...
+                </div>
+              </div>
+            )}
+          </div>
+        ) : (
+          <ul class='atom-stepper'>
+            {this.parsedSteps.map(({ title, completed }, index) => {
+              const isActive = index === this.activeStep
+              const shouldDisabled =
+                this.disabledUnrenderedStep && index > this.activeStep
```
tassioFront, line 84: Question: Should the icon button be interactive when disabledUnrenderedStep is false?

packages/core/src/components.d.ts
```diff
@@ -491,6 +491,28 @@ export namespace Components {
...
+        /**
+         * @deprecated use camelCase instead. Support for dash-casing will be removed in Stencil v5.
+         */
+        "step-title"?: string;
+        "steps": | string
+    | {
+        title: string
+        completed: boolean
+      }[];
+    }
```
luiz-ortega, line 515: suggestion/question: we really need the `active-step` prop? Couldn't we get the active step by referencing the `completed` key on the `steps` prop? A more simplified interface is better

packages/core/src/components/stepper/stepper.tsx
igorwessel: @gabrielduete if it is for testing, when the prop is objects/etc we must set it in the instance itself. we don't need to do this conversion here.
```
// For Test
document.querySelector('atom-stepper').steps = [{}]
```

packages/core/src/components/stepper/stepper.scss
gabrielduete: Nice!

Made: https://github.com/juntossomosmais/atomium/pull/726/commits/be7d94e5ab5adf29c1017260d4ebd1c73c4c2540

packages/core/src/components/stepper/stepper.tsx
gabrielduete: Nice! I chose to use `addCompletedClass`

Made: https://github.com/juntossomosmais/atomium/pull/726/commits/8e6b6de2af6f8c8dc7cedbd93d8d154026d0aa0d

gabrielduete: Nice! I chose to use `addCompletedClass`

https://github.com/juntossomosmais/atomium/pull/726/commits/8e6b6de2af6f8c8dc7cedbd93d8d154026d0aa0d

packages/core/src/components/stepper/stepper.tsx
gabrielduete: When true.

disabledUnrenderedStep is a prop that disables all steps after the current step (index > this.activeStep). But this is an optional feature, so the prop must be true, and the index must be greater than the current step.

image

packages/core/src/components.d.ts
gabrielduete: Hmm, I think this would pollute the steps dictionary a bit, which would make it harder to manipulate the object in the parent component where the stepper will be used.

Keeping an actualStep prop ensures that it will only change the state of the current step, without needing to interact with the object again.

packages/core/src/components/stepper/stepper.scss
```diff
@@ -0,0 +1,118 @@
...
+    align-items: center;
+    flex-direction: column;
+    height: 100%;
+    justify-content: center;
+
+    &:not(:last-child)::after {
+      height: 1px;
+      left: 60%;
+      top: var(--spacing-small);
+      width: 80%;
```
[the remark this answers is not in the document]
gabrielduete: Yeah!

Made: https://github.com/juntossomosmais/atomium/pull/726/commits/7a3baf7facd6e5da05cfed011e64d48e4ce51505

packages/core/src/components/stepper/stepper.tsx
tassioFront: Thanks for explaining, @gabrielduete. Guess I didn't make myself clear; I mean, should the icon button be **clickable** (click event to do something like go to next step) when disabledUnrenderedStep is `false`? If not, what is the diff between them?

tassioFront: Hey man, I saw the Figma. I guess we don't need this prop, as it looks like we shouldn't have a disabledUnrenderedStep=false scenario. Did I get it right?

gabrielduete: > Thanks for explaining, @gabrielduete. Guess I didn't make myself clear; I mean, should the icon button be clickable (click event to do something like go to next step) when disabledUnrenderedStep is false? If not, what is the diff between them?

Yeah! The difference is only visual, it has those “disabled” styles that we use in components. The option to disable the button depends on the logic of the page that will use the component.

> Hey man, I saw the Figma. I guess we don't need this prop, as it looks like we shouldn't have a disabledUnrenderedStep=false scenario. Did I get it right?

Talking to the UX team, we saw that this scenario does not exist today, but it would be good to leave it in case we need it.

gabrielduete: > Thanks for explaining, @gabrielduete. Guess I didn't make myself clear; I mean, should the icon button be clickable (click event to do something like go to next step) when disabledUnrenderedStep is false? If not, what is the diff between them?

Yeah! The difference is only visual, it has those “disabled” styles that we use in components. The option to disable the button depends on the logic of the page that will use the component.

> Hey man, I saw the Figma. I guess we don't need this prop, as it looks like we shouldn't have a disabledUnrenderedStep=false scenario. Did I get it right?

Talking to the UX team, we saw that this scenario does not exist today, but it would be good to keep it in case we need it.

tassioFront: ok man! Thanks

tassioFront (approved):

packages/core/src/components/stepper/stepper.tsx
gabrielduete: By way of props, but I had to do this conversion, specificallyfor unit tests

Stencil doesn't automatically parse objects, the component expects an array of objects, but Stencil props are strings in HTML.

So I had to manually parse the received JSON. =/

packages/core/src/components/stepper/stepper.tsx
gabrielduete: > Why do you have to do this?

When I had to [convert to JSON to fix the tests](https://github.com/juntossomosmais/atomium/pull/726#discussion_r2089919951), I noticed that the component was not reactive, because I couldn't understand that the object was changing, so I had to leave it hidden in the Watch

> if it is for testing, when the prop is objects/etc we must set it in the instance itself. we don't need to do this conversion here.

I tried this initially, but I was unsuccessful. Using it this way, debugging, I saw that the object was always empty =/, so I had to [manually parse it to pass the tests](https://github.com/juntossomosmais/atomium/pull/726#discussion_r2089919951), because of this, it was necessary to use watch to make the component reactive.

igorwessel: I believe your test was missing waiting for changes in the DOM, I did a test and everything went ok passing the props directly through the instance.
```js
  it('should render completed steps', async () => {
    const page = await newSpecPage({
      components: [AtomStepper],
      html: `<atom-stepper active-step="1" role="stepper"></atom-stepper>`,
    })

    page.root?.steps = [
      { title: 'Step 1', completed: true },
      { title: 'Step 2', completed: false },
    ]

    await page.waitForChanges()

    expect(page.root).toEqualHtml(`
      <atom-stepper active-step="1" role="stepper">
        <mock:shadow-root>
          <ul class="atom-stepper">
            <li class="is-completed is-last step">
              <div class="number">
                <atom-icon icon="check" color="white" size="16"></atom-icon>
              </div>
              <span class="title">Step 1</span>
            </li>
            <li class="step">
              <div class="number">2</div>
              <span class="is-active title">Step 2</span>
            </li>
          </ul>
        </mock:shadow-root>
      </atom-stepper>
    `)
  })
```

igorwessel (dismissed):

packages/core/src/components/stepper/stepper.tsx
gabrielduete: Very good, @igorwessel! I had to use `rootInstance` and it worked. Thanks, :mage-wessel: 🧙🏽‍♂️

@MayaraRMA , With this, i remove the parsing! /o/

Commit: https://github.com/juntossomosmais/atomium/pull/726/commits/da9bfe83eea84e5e35277ef3a5dedc823d3c3e71

packages/core/src/components/stepper/stepper.tsx
gabrielduete: Removed: https://github.com/juntossomosmais/atomium/pull/726#discussion_r2091373203

PR #759, opened 2025-06-02
fix(core): include core css on exports

matheus-rosa-jsm:
[Task](https://juntossomosmais.monday.com/boards/5143488854/views/113762127/pulses/9286490020)

## Summary

After add explicit exports, that will be needed after stencil upgrade, core.css is not found for imports.

## Evidences

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -56,6 +56,10 @@
     "./vue": {
       "types": "./vue/dist/types/index.d.ts",
       "import": "./vue/dist/esm/index.js"
+    },
+    "./core.css": {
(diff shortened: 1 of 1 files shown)


packages/core/package.json
```diff
@@ -56,6 +56,10 @@
     "./vue": {
       "types": "./vue/dist/types/index.d.ts",
       "import": "./vue/dist/esm/index.js"
```
igorwessel, line 58: **question:** ‏vue don't have cjs build?

anabrook (approved):

PR #818, opened 2026-04-28
feat(tooltip): add fixed prop to keep tooltip persistently open

luiz-ortega:
## Summary

- Adds `fixed` prop to `atom-tooltip` that prevents all auto-close triggers (mouseleave, blur, outside click)
- When `fixed=true`, only the X button can dismiss the tooltip
- Close button is now also rendered when `fixed=true` (previously only on `action=click` or mobile)

## Changes

- `tooltip.tsx`: new `fixed` prop; `hide()` bails when `fixed`; `forceHide()` for X button; shared `_doHide()` helper
- `tooltip.spec.ts`: 3 new tests covering fixed behavior
- `tooltip.args.ts`: `fixed` argType + default
- All 3 story files: `fixed` prop wired + `Fixed` story added

## Test plan

- [ ] `npm test` passes
- [ ] Storybook `Fixed` story shows tooltip open with X button
- [ ] Hovering off / clicking outside does not close fixed tooltip
- [ ] Clicking X closes fixed tooltip

### DIFF
--- a/packages/core/src/components/tooltip/tooltip.scss
+++ b/packages/core/src/components/tooltip/tooltip.scss
@@ -20,7 +20,7 @@ $placements: 'top', 'bottom', 'left', 'right';
 
     background: var(--background, var(--color-neutral-light-1));
     border-radius: var(--border-radius);
-    color: var(--color-neutral-white);
+    color: var(--color, var(--color-neutral-white));
     font: var(--text-body-small);
     letter-spacing: var(--text-body-small-letter);
     padding: var(--spacing-xsmall);
--- a/packages/core/src/components/tooltip/tooltip.spec.ts
+++ b/packages/core/src/components/tooltip/tooltip.spec.ts
@@ -194,4 +194,62 @@ describe('AtomTooltip', () => {
         ?.shadowRoot?.querySelector('.atom-tooltip')
     ).toHaveAttribute('data-show')
   })
+
+  describe('fixed prop', () => {
+    const fixedTooltip = `
+      <button id="fixed" aria-describedby="fixed--tooltip">Trigger</button>
+      <atom-tooltip id="fixed--tooltip" element="fixed" open="true" action="hover" fixed="true" part="tooltip">Content</atom-tooltip>
+    `
+
+    it('should stay open after mouseleave when fixed', async () => {
+      const page = await newSpecPage({
+        components: [AtomTooltip],
+        html: fixedTooltip,
+      })
+
+      const element = page.body.querySelector('#fixed')
+
+      element?.dispatchEvent(new Event('mouseleave'))
+
+      await page.waitForChanges()
+
+      expect(
+        page.root?.shadowRoot?.querySelector('.atom-tooltip')
+      ).toHaveAttribute('data-show')
+    })
+
+    it('should stay open after outside click when fixed', async () => {
+      const page = await newSpecPage({
+        components: [AtomTooltip],
+        html: fixedTooltip,
+      })
+
+      page.body.dispatchEvent(new Event('click', { bubbles: true }))
+
+      await page.waitForChanges()
+
+      expect(
+        page.root?.shadowRoot?.querySelector('.atom-tooltip')
+      ).toHaveAttribute('data-show')
+    })
+
+    it('should close when close button is clicked when fixed', async () => {
+      const page = await newSpecPage({
+        components: [AtomTooltip],
+        html: fixedTooltip,
+      })
+
+      const closeButton = page.root?.shadowRoot?.querySelector(
+        '.atom-tooltip__action--close'
+      ) as HTMLElement
+
+      closeButton?.click()
+
+      await page.waitForChanges()
+
+      expect(
+        page.root?.shadowRoot?.querySelector('.atom-tooltip')
+      ).toHaveAttribute('data-hide')
+    })
+  })
 })
--- a/packages/core/src/components/tooltip/tooltip.tsx
+++ b/packages/core/src/components/tooltip/tooltip.tsx
@@ -62,6 +62,8 @@ export class AtomTooltip {
 
   @Prop() action: 'hover' | 'click' = 'hover'
 
+  @Prop({ mutable: true }) fixed = false
+
   @Watch('action')
   updateEvents() {
     this.untachEvents()
@@ -218,6 +220,16 @@ export class AtomTooltip {
   }
 
   private readonly hide = () => {
+    if (this.fixed) return
+
+    this._doHide()
+  }
+
+  private readonly forceHide = () => {
+    this._doHide()
+  }
+
+  private readonly _doHide = () => {
     this.open = false
     this.atomClose.emit()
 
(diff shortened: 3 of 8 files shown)


packages/core/src/components/tooltip/tooltip.scss
```diff
@@ -20,7 +20,7 @@ $placements: 'top', 'bottom', 'left', 'right';
 
     background: var(--background, var(--color-neutral-light-1));
     border-radius: var(--border-radius);
-    color: var(--color-neutral-white);
+    color: var(--color, var(--color-neutral-white));
```
igorwessel, line 23: **question:** ‏why `--color`?

luiz-ortega: @igorwessel since we are enabling customizations for the background we should be able to customize the color from the text to achieve a clean readable contrast

felipefialho (changes requested):
## Summary

- `@Prop({ mutable: true })` should be plain `@Prop()` for `fixed` — config flags must not be mutable by the component
- `forceHide` is an unnecessary wrapper around `_doHide` — remove it and reference `_doHide` directly from JSX
- Keyboard users cannot dismiss a `fixed` tooltip via Escape — WCAG 1.4.13 violation requiring immediate fix
- Missing test: `blur` event suppression in fixed mode (both `mouseleave` and `blur` are in `_eventsToHideHover`)
- Missing test: close button is actually rendered when `fixed=true` + `action=hover`

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -62,6 +62,8 @@ export class AtomTooltip {
 
   @Prop() action: 'hover' | 'click' = 'hover'
 
+  @Prop({ mutable: true }) fixed = false
```
felipefialho, line 65: [issue] `@Prop({ mutable: true })` is wrong for `fixed`. This is a configuration flag passed by the consumer — the component never writes back to it, so marking it mutable is semantically incorrect and misleads Stencil's data-flow model. Use a plain `@Prop()`:

```ts
@Prop() fixed = false
```

As a side-effect, fixing this will also regenerate `components.d.ts` correctly — currently `fixed` is non-optional in `AtomTooltipAttributes` but optional everywhere else, which is inconsistent.

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -218,6 +220,16 @@ export class AtomTooltip {
   }
 
   private readonly hide = () => {
+    if (this.fixed) return
+
+    this._doHide()
+  }
+
+  private readonly forceHide = () => {
```
felipefialho, line 228: [issue] `forceHide` is a one-line wrapper with no added logic — its sole body is `this._doHide()`. This adds an extra name to track for no benefit. Reference `_doHide` directly from the JSX `onClick` and remove `forceHide` entirely:

```tsx
onClick={this._doHide}
```

If the convention in this codebase is that `_`-prefixed methods must not appear in JSX, then keep `forceHide` and move the implementation there, removing `_doHide` as the indirection layer instead.

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -248,11 +260,11 @@ export class AtomTooltip {
           <div class='atom-tooltip__content'>
             <slot />
 
-            {(this.action === 'click' || isMobile()) && (
+            {(this.action === 'click' || isMobile() || this.fixed) && (
```
felipefialho, line 263: [issue] Keyboard accessibility gap: with `fixed=true`, the Escape key no longer closes the tooltip because `hide()` bails early unconditionally. WCAG 1.4.13 requires that non-modal content be dismissible by keyboard. This blocks keyboard-only and screen-reader users.

Add a `@Listen('keydown', { target: 'window' })` handler that routes Escape to `forceHide` regardless of `fixed`:

```ts
@Listen('keydown', { target: 'window' })
handleKeydown(ev: KeyboardEvent) {
  if (ev.key === 'Escape' && this.open) this.forceHide()
}
```

Also: the close button's `aria-label` is currently `'Fechar'` (Portuguese). For an international component library, consider whether it should be localisable or use a language-neutral `aria-label`.

packages/core/src/components/tooltip/tooltip.spec.ts
```diff
@@ -194,4 +194,62 @@ describe('AtomTooltip', () => {
         ?.shadowRoot?.querySelector('.atom-tooltip')
     ).toHaveAttribute('data-show')
   })
+
```
felipefialho, line 197: [suggestion] The existing spec already has a `createTooltip(id, action, open)` factory at the top of the describe block. This `fixedTooltip` inline HTML string duplicates the same structure. Extend the factory to accept an optional `fixed` parameter instead of maintaining a parallel template:

```ts
const createTooltip = (
  id: string,
  action: AtomTooltip['action'] = 'hover',
  open = false,
  fixed = false
) => `
  <button id="${id}" aria-describedby="${id}--tooltip">Hover</button>
  <atom-tooltip id="${id}--tooltip" element="${id}" open="${open}" action="${action}" fixed="${fixed}" part="tooltip">John Doe</atom-tooltip>
`

// Then:
const fixedTooltip = createTooltip('fixed', 'hover', true, true)
```

This avoids the two templates drifting apart if the component's HTML shape changes.

packages/core/src/components/tooltip/tooltip.spec.ts
```diff
@@ -194,4 +194,62 @@ describe('AtomTooltip', () => {
...
+  describe('fixed prop', () => {
+    const fixedTooltip = `
+      <button id="fixed" aria-describedby="fixed--tooltip">Trigger</button>
+      <atom-tooltip id="fixed--tooltip" element="fixed" open="true" action="hover" fixed="true" part="tooltip">Content</atom-tooltip>
+    `
```
felipefialho, line 202: [issue] The existing test suite covers both `mouseleave` and `blur` as hide triggers via `it.each(['mouseleave', 'blur'])`. This `fixed` block only covers `mouseleave`. A `blur` event on the trigger should equally be suppressed — add a parallel test:

```ts
it('should stay open after blur when fixed', async () => {
  const page = await newSpecPage({ components: [AtomTooltip], html: fixedTooltip })
  page.body.querySelector('#fixed')?.dispatchEvent(new Event('blur'))
  await page.waitForChanges()
  expect(page.root?.shadowRoot?.querySelector('.atom-tooltip')).toHaveAttribute('data-show')
})
```

Also missing: a test asserting the close button is **rendered** (not just functional) when `fixed=true` + `action=hover`. If the render condition regressed, the current close-button click test would silently pass because clicking a non-existent element is a no-op.

packages/core/src/components/tooltip/tooltip.scss
felipefialho, line 23: [suggestion] `--color` is a very common CSS custom property name — any ancestor in the host page that sets it will bleed into the tooltip text. Namespace it to the component, consistent with how `--background` and `--border-radius` are typically scoped in Stencil design systems:

```scss
color: var(--atom-tooltip-color, var(--color-neutral-white));
```

This follows the same pattern as the `--background` fallback already on the line above.

packages/core/src/components/tooltip/tooltip.spec.ts
```diff
@@ -194,4 +194,62 @@ describe('AtomTooltip', () => {
         ?.shadowRoot?.querySelector('.atom-tooltip')
     ).toHaveAttribute('data-show')
   })
```
felipefialho, line 196: [praise] Clean test organisation: isolating all `fixed` behaviour in its own `describe` block with a shared HTML fixture keeps the new tests cohesive and easy to expand. The three cases cover the main user-facing contract (stay open on external events, close only via button).

PR #821, opened 2026-05-06
fix(core): prevent ssr crash in request submit polyfill

dyhalmeida:
[Task](https://juntossomosmais.monday.com/boards/5337789717/pulses/11949610815)

## Summary

Fix a crash in the `requestSubmit` polyfill (`packages/core/src/polyfills/form-request-submit.js`)
that caused Next.js SSR builds to fail with:
`TypeError: Cannot read properties of undefined (reading 'prototype')`

**Root cause:** The IIFE argument `globalThis.HTMLFormElement.prototype` was evaluated
eagerly in Node.js environments where `HTMLFormElement` does not exist (DOM-only API).
The `&&` short-circuit evaluation returned `true`, which then tried to access `.prototype`
on `undefined`, crashing before the internal `if (!prototype) return` guard could run.

**Fix:** Added optional chaining (`?.`) so `globalThis.HTMLFormElement?.prototype`
returns `undefined` instead of throwing, allowing the guard to exit cleanly.

- **Browser impact:** None — `HTMLFormElement` always exists in browsers.
- **SSR/Node.js impact:** Polyfill is now safely skipped via early return.

### DIFF
--- a/packages/core/src/polyfills/form-request-submit.js
+++ b/packages/core/src/polyfills/form-request-submit.js
@@ -62,4 +62,4 @@
       name
     )
   }
-})(typeof globalThis !== 'undefined' && globalThis.HTMLFormElement.prototype)
+})(typeof globalThis !== 'undefined' && globalThis.HTMLFormElement?.prototype)
--- a/packages/core/src/polyfills/form-request-submit.spec.js
+++ b/packages/core/src/polyfills/form-request-submit.spec.js
@@ -1,0 +1,15 @@
+describe('form-request-submit polyfill', () => {
+  it('should not throw when HTMLFormElement is undefined (SSR environment)', () => {
+    const original = globalThis.HTMLFormElement
+
+    delete globalThis.HTMLFormElement
+
+    expect(() => {
+      jest.isolateModules(() => {
+        require('./form-request-submit.js')
+      })
+    }).not.toThrow()
+
+    globalThis.HTMLFormElement = original
+  })
+})


felipefialho (commented):
## Code Review

**Positive Points**

The fix is minimal, surgical, and correct. The PR description is excellent: it clearly explains the root cause (eager IIFE argument evaluation in Node.js), the mechanism of failure (`&&` short-circuit returning `true` before the guard runs), and the impact on both browser and SSR environments. Optional chaining is exactly the right idiom here.

---

**Findings**

- [MEDIUM] No test coverage for the new SSR execution path introduced by this fix. The existing `button.spec.ts` covers the polyfill in jsdom (where `HTMLFormElement` always exists), but the specific scenario where `globalThis.HTMLFormElement` is `undefined` (the crash this PR fixes) is not verified by any test.

packages/core/src/polyfills/form-request-submit.js
```diff
@@ -62,4 +62,4 @@
       name
     )
   }
-})(typeof globalThis !== 'undefined' && globalThis.HTMLFormElement.prototype)
+})(typeof globalThis !== 'undefined' && globalThis.HTMLFormElement?.prototype)
```
felipefialho, line 65: [suggestion] The SSR crash path fixed here has no test coverage. Consider adding a test that simulates `globalThis.HTMLFormElement` being `undefined` and asserts the module loads without throwing:

```js
it('should not throw when HTMLFormElement is undefined (SSR environment)', async () => {
  const original = globalThis.HTMLFormElement
  delete globalThis.HTMLFormElement

  await expect(
    jest.isolateModulesAsync(() => import('../../polyfills/form-request-submit.js'))
  ).resolves.not.toThrow()

  globalThis.HTMLFormElement = original
})
```

This guards against regressions if the IIFE argument is ever changed again.

dyhalmeida: done 0d4a6b994

mauriciomutte (approved):

gabrielduete (approved):

packages/core/src/polyfills/form-request-submit.js
felipefialho: Looks good, thanks!

felipefialho (approved):
All issues resolved. The SSR guard test was added and correctly isolates the module during require. 

[praise] Good use of `jest.isolateModules` + `require` — simpler and more idiomatic than the async variant I suggested.

[nit] One minor hardening suggestion for the cleanup: if the assertion ever fails, `globalThis.HTMLFormElement = original` won't execute (Jest throws before reaching it). A `try/finally` or moving the restore into `afterEach` would make this bulletproof — but it's non-blocking.

PR #829, opened 2026-05-28
chore(eslint): migrate to flat config and eslint v10

felipefialho:
## Why

Following the `@juntossomosmais/linters@0.25.1` release that adopts the flat config and ESLint v10. atomium was one of the consumers blocked on Phase 2 of the rollout. With `linters@0.25.1` (which swaps `eslint-plugin-import` for `eslint-plugin-import-x` to work around ESLint v10's removal of `SourceCode#getTokenOrCommentAfter`), atomium can now move off the legacy eslintrc setup.

### Tooling

- Bump `eslint` to `^10` and `@juntossomosmais/linters` to `^0.25.1`.
- Add the deps the new flat base needs (`@eslint/js`, `eslint-config-prettier`, `eslint-plugin-import-x`, `eslint-plugin-react`, `eslint-plugin-react-hooks`, `globals`, `typescript-eslint`).
- Drop `eslint-plugin-import` (legacy CJS) in favor of `eslint-plugin-import-x`.
- Scope the `brace-expansion` CVE override to the v1 line and bump the v2 line to `2.1.1`, otherwise the new `minimatch` cannot resolve its dependency.

### Config

- Replace `.eslintrc.js` + four per-package `.eslintrc.json` files with a single `eslint.config.mjs` at the root.
- The new config imports the flat react base from `@juntossomosmais/linters/eslint-config/flat/react.mjs`, applies `eslint-plugin-storybook`'s `flat/recommended` config, and restores the project overrides (spec, stories, scripts, Stencil components).
- Stencil tsx files now also disable `react/react-in-jsx-scope` and accept `innerHTML` as a known property.
- Story files allow `no-console`, `@typescript-eslint/no-explicit-any`, `sonarjs/slow-regex`, and `sonarjs/no-nested-assignment` (these are example/teaching patterns, not production code).
- Storybook config files accept `@typescript-eslint/no-explicit-any` (required by framework typings).
- Unused arguments prefixed with `_` are now accepted (conventional pattern, kept the existing Stencil `^h$` ignore).

### Autofix output

`eslint --fix` reordered imports and inserted padding lines across stories, `.storybook` configs, and tokens scripts. Two code-level fixes:

- `textarea.{react,vue}.stories.tsx`: `maxLength` → `_maxLength` (unused arg).
- `tooltip.tsx`: `aria-hidden` → `aria-hidden={true}` (boolean prop, per `react/jsx-boolean-value`).

## Verification

```bash
npx eslint .
# clean: 0 errors, 0 warnings
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mauriciomutte (dismissed):

eslint.config.mjs
```diff
@@ -0,0 +1,102 @@
...
+    // scope for JSX.
+    files: ['packages/core/src/components/**/*.tsx'],
+    rules: {
+      'react/no-unknown-property': [
+        'error',
+        { ignore: ['class', 'innerHTML'] },
+      ],
+      'react/react-in-jsx-scope': 'off',
+    },
+  },
```
igorwessel: suggestion: maybe it would make sense to scope the React preset only to the packages that actually use React, instead of applying ...react globally.

Something like:
```js
...react.map((config) => ({
  ...config,
  files: ['packages/react-app/**/*.{js,jsx,ts,tsx}'],
}))
```

Or using basePath, which might be even clearer in a monorepo context:
```js
{
  basePath: 'packages/react-app',
  extends: react,
}
```

This would avoid loading React-specific rules/plugins for Stencil/Vue/etc packages.

felipefialho: @igorwessel Good point in general! Re-checked the lint surface here: `packages/react/**` (and `packages/vue/**`, plus the generated `packages/core/{react,vue,loader,output-target}/**`) are all in the `ignores` list at the top of `eslint.config.mjs`, so the only files actually linted by this config are the Stencil TSX components under `packages/core/src/**` plus root-level scripts and stories. Scoping `...react` to `packages/react/**` would point it at a fully-ignored path. The global spread is acting mostly as "JSX parsing + sensible defaults" for the Stencil TSX files (which the override below narrows further for Stencil-specific quirks like `class` and `innerHTML`). If `packages/react/**` is ever un-ignored, then the scoping you describe would absolutely be the right move, happy to revisit then.

igorwessel (dismissed):

eslint.config.mjs
felipefialho: @igorwessel if the lint surface is Stencil-only, then even the react preset itself earns nothing here. Switched the config to use just `base` in [fdf9e04](https://github.com/juntossomosmais/atomium/commit/fdf9e041d). The base already enables JSX parsing via `parserOptions.ecmaFeatures.jsx`, so dropping `...react` lets us also remove the Stencil-only override block (`react/no-unknown-property`, `react/react-in-jsx-scope`) and drop `eslint-plugin-react` + `eslint-plugin-react-hooks` from devDependencies. Lint stays clean (0 errors, 0 warnings).

igorwessel (approved):
lgtm

mauriciomutte (approved):

PR #855, opened 2026-06-17
feat(core): add disable close button option to modal

brunosf:
[Task](https://juntossomosmais.monday.com/boards/5143488854/views/113762127/pulses/12169528213?userId=30945655)

## Summary

Adds a new `disableCloseButton` prop to the `atom-modal` component. When set to `true`:

- The modal's close button is rendered in a `disabled` state.
- Clicking the close button no longer emits the `atomCloseClick` event nor dismisses the modal (the `handleCloseClick` handler returns early).

The prop defaults to `false`, so existing behavior is unchanged. The new option is also exposed in the Storybook controls/args under the component properties category.

Covered by two new unit tests:
- Verifies the close button renders with the `disabled` attribute when `disableCloseButton` is `true`.
- Verifies `atomCloseClick` is not emitted and `dismiss()` is not called on click when `disableCloseButton` is `true`.

## Evidences

_N/A — internal behavior/prop change. See unit tests and Storybook control for verification._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/core/src/components/modal/modal.spec.ts
+++ b/packages/core/src/components/modal/modal.spec.ts
@@ -160,6 +160,45 @@ describe('atom-modal', () => {
     expect(spySecondary).toHaveBeenCalled()
   })
 
+  it('should render the close button as disabled when disableCloseButton is true', async () => {
+    page = await newSpecPage({
+      components: [AtomModal],
+      html: `
+      <atom-modal is-open="true" disable-close-button="true">
+        Modal content
+      </atom-modal>
+    `,
+    })
+
+    const closeButton = page.root?.querySelector('.atom-modal__close')
+
+    expect(closeButton?.hasAttribute('disabled')).toBe(true)
+  })
+
+  it('should not emit atomCloseClick nor dismiss when disableCloseButton is true', async () => {
+    page = await newSpecPage({
+      components: [AtomModal],
+      html: `
+      <atom-modal is-open="true" disable-close-button="true">
+        Modal content
+      </atom-modal>
+    `,
+    })
+
+    const dismiss = jest.fn()
+    const spyClose = jest.fn()
+
+    page.rootInstance.modal = { dismiss, close: jest.fn() }
+    page.root?.addEventListener('atomCloseClick', spyClose)
+
+    const closeButton = page.root?.querySelector('.atom-modal__close')
+
+    closeButton?.dispatchEvent(new Event('click'))
+
+    expect(spyClose).not.toHaveBeenCalled()
+    expect(dismiss).not.toHaveBeenCalled()
+  })
+
   it('should render progress bar when progress is passed even if it is zero', async () => {
     await page.setContent(`
       <atom-modal>
--- a/packages/core/src/components/modal/modal.tsx
+++ b/packages/core/src/components/modal/modal.tsx
@@ -34,6 +34,7 @@ export class AtomModal {
   @Prop() hasFooter = true
   @Prop() disablePrimaryButton = false
   @Prop() disableSecondaryButton = false
+  @Prop() disableCloseButton = false
   @Prop({ mutable: true }) isOpen = false
   @Prop({ mutable: true }) canDismiss?: boolean
   @Prop() metaData?: MetaData = {}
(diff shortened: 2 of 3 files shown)


felipefialho (changes requested):
## Summary

- **Pattern deviation:** `handleCloseClick` has an early-return guard for `disableCloseButton`, but the sibling handlers (`handlePrimaryClick`, `handleSecondaryClick`) have no such guard for their respective disable props — they rely solely on `atom-button disabled`. The guard is redundant and breaks the internal consistency of the component.
- **Test reliability:** Test 2 dispatches `new Event('click')` directly on the `atom-button` host element in Stencil's spec-page environment, which does not reach the `onClick` JSX handler (`handleCloseClick`). The assertions pass for reasons unrelated to the guard, giving false confidence that the behavior is validated.

packages/core/src/components/modal/modal.tsx
```diff
@@ -74,6 +75,10 @@ export class AtomModal {
   }
 
   private readonly handleCloseClick = async () => {
+    if (this.disableCloseButton) {
```
felipefialho, line 78: [issue] This early-return guard is inconsistent with the sibling disable props. Neither `handlePrimaryClick` nor `handleSecondaryClick` has a guard for their respective `disablePrimary/SecondaryButton` props — they delegate that responsibility entirely to `disabled={...}` on `<atom-button>`. When `disabled` is set, the browser (and Stencil's scoped component model) prevents the click from reaching `onClick`, making this guard redundant.

Remove the guard and let the `disabled` prop on the button (line 137) be the sole enforcement mechanism, consistent with the established pattern:

```ts
private readonly handleCloseClick = async () => {
  this.atomCloseClick.emit(this.modal)
  this.modal.dismiss()
}
```

packages/core/src/components/modal/modal.spec.ts
```diff
@@ -160,6 +160,45 @@ describe('atom-modal', () => {
...
+
+    const dismiss = jest.fn()
+    const spyClose = jest.fn()
+
+    page.rootInstance.modal = { dismiss, close: jest.fn() }
+    page.root?.addEventListener('atomCloseClick', spyClose)
+
+    const closeButton = page.root?.querySelector('.atom-modal__close')
+
+    closeButton?.dispatchEvent(new Event('click'))
```
felipefialho, line 196: [issue] This `dispatchEvent(new Event('click'))` targets the `atom-button` custom element host, not its internal shadow or scoped button. In Stencil's spec-page (JSDOM) environment, this synthetic event does not bubble through the component's `onClick` JSX prop binding — `handleCloseClick` is never called. The test assertions (`not.toHaveBeenCalled`) pass vacuously: both spy and dismiss would be uncalled regardless of whether the guard exists.

To test the guard meaningfully, call the handler directly on the instance:

```ts
await page.rootInstance.handleCloseClick()

expect(spyClose).not.toHaveBeenCalled()
expect(dismiss).not.toHaveBeenCalled()
```

Note: if the guard is removed (per the comment on `modal.tsx`), this test becomes unnecessary — the disabled-attribute test (line 163) is the correct coverage for this prop.

packages/core/src/components/modal/modal.spec.ts
```diff
@@ -160,6 +160,45 @@ describe('atom-modal', () => {
     expect(spySecondary).toHaveBeenCalled()
   })
 
+  it('should render the close button as disabled when disableCloseButton is true', async () => {
```
felipefialho, line 163: [praise] This test is the right approach: render the component with `disable-close-button="true"` via the HTML API and assert the resulting `disabled` attribute is present. This matches exactly how `disablePrimaryButton` and `disableSecondaryButton` are tested in the existing spec.

packages/core/src/components/modal/modal.tsx
```diff
@@ -129,6 +134,7 @@ export class AtomModal {
               color='secondary'
               fill='clear'
               shape='circle'
+              disabled={this.disableCloseButton}
```
felipefialho, line 137: [thought] Worth a team discussion: native `disabled` removes the button from the tab order and can cause screen readers to silently skip it, giving users with assistive technology no indication that closing is currently unavailable. The existing `disablePrimary/SecondaryButton` props have the same behavior, so this PR is consistent with the current approach. If the pattern is ever revisited, `aria-disabled="true"` (with the JS guard to block the action) would keep the button focusable and announceable. Not a blocker for this PR.

PR #861, opened 2026-07-07
fix(modal): cap height and pin header/footer on tall content

igorwessel:
## Summary

`ion-modal::part(content)` (the actual dialog box Ionic renders) had `height: auto` with no `max-height`. On tablet/desktop, a modal whose content is taller than the viewport just overflows past both edges instead of scrolling internally - the header gets pushed above the top of the screen and the footer below the bottom, both effectively invisible.

Fix: cap `part(content)` at `max-height: 90vh` and make it the scroll container (`overflow-y: auto`), then pin `.atom-modal__header`/`.atom-modal__footer` with `position: sticky` so they stay visible while the middle content scrolls underneath them.

This intentionally does **not** touch Ionic's own internal `.ion-page` delegate-host wrapper or `.atom-modal__content`'s sizing - an earlier attempt using `display: flex` on the wrapper needed to fight `.ion-page`'s `height: 100%` (which resolves to `auto` against an `auto`-height ancestor even once `max-height` clamps it - a CSS percentage-vs-auto-parent quirk). `position: sticky` sidesteps that entirely: it only needs a scrolling ancestor, regardless of how intervening elements resolve their own height.

Scoped to `@include above('small')` only - the existing mobile fullscreen modal (`@include below('small')`, `height: 100%`) is untouched and doesn't have this bug (its containing block already has a definite height).

## Evidences

No visual snapshot from this environment. Verified computed layout with Playwright against the real compiled `ion-modal`/`atom-modal` web components (not jsdom, which doesn't do real CSS layout) at multiple viewports:

- Desktop, short viewport (800px) + content taller than the viewport: before the fix, header top was at `-73px` (above viewport) and footer bottom at `873px` (below viewport) - both invisible. After the fix, both stay fully within the viewport at every scroll position, including scrolled to the max (`scrollTop` at its maximum), and the content between them gets a real scrollbar.
- Desktop, tall viewport + short content: modal still shrinks to fit content (height well under the `90vh` cap) - no regression for the common case.
- Mobile viewport (<768px): `max-height: none`, `overflow: hidden`, `height` = 100% of viewport, no `position: sticky` applied - fullscreen behavior untouched.

### DIFF
--- a/packages/core/src/components/modal/modal.scss
+++ b/packages/core/src/components/modal/modal.scss
@@ -25,6 +25,13 @@
     padding: var(--spacing-large);
     padding-top: 0;
 
+    // Content is the only scroll container once ion-modal::part(content) is capped (see below);
+    // keeping this as auto would silently activate a nested scrollbar if this element ever gets
+    // an explicit height.
+    @include above('small') {
+      overflow-y: visible;
+    }
+
     &--divided {
       border-bottom: 1px solid var(--color-neutral-light-3);
       border-top: 1px solid var(--color-neutral-light-3);
@@ -41,7 +48,7 @@
 
     // Pins the footer while the content below scrolls (see ion-modal::part(content)).
     @include above('small') {
-      background: var(--background);
+      background: var(--background, var(--color-neutral-white));
       bottom: 0;
       position: sticky;
       z-index: var(--zindex-1);
@@ -56,7 +63,7 @@
 
     // Pins the header while the content below scrolls (see ion-modal::part(content)).
     @include above('small') {
-      background: var(--background);
+      background: var(--background, var(--color-neutral-white));
       position: sticky;
       top: 0;
       z-index: var(--zindex-1);
@@ -114,6 +121,10 @@ ion-modal::part(content) {
   @include above('small') {
     max-height: 90vh;
     overflow-y: auto;
+
+    // Reserves space so a keyboard-focused element near the top/bottom isn't hidden behind the
+    // sticky header/footer, which native focus scrolling doesn't otherwise account for.
+    scroll-padding-block: var(--spacing-xxxxlarge);
   }
 
   @include below('small') {
--- a/packages/core/src/components/modal/modal.scss
+++ b/packages/core/src/components/modal/modal.scss
@@ -38,6 +38,14 @@
     gap: var(--spacing-medium);
     justify-content: flex-end;
     padding: var(--spacing-large);
+
+    // Pins the footer while the content below scrolls (see ion-modal::part(content)).
+    @include above('small') {
+      background: var(--background);
+      bottom: 0;
+      position: sticky;
+      z-index: var(--zindex-1);
+    }
   }
 
   &__header {
@@ -45,6 +53,14 @@
     padding: var(--spacing-large);
     padding-right: var(--atom-modal-padding-header);
     position: relative;
+
+    // Pins the header while the content below scrolls (see ion-modal::part(content)).
+    @include above('small') {
+      background: var(--background);
+      position: sticky;
+      top: 0;
+      z-index: var(--zindex-1);
+    }
   }
 
   &__icon-type {
@@ -91,6 +107,14 @@
 ion-modal::part(content) {
   --border-radius: 8px;
   height: auto;
+
+  // With no max-height, content taller than the viewport just overflows past both edges instead
+  // of scrolling internally, pushing the header/footer off-screen. Capping it and letting it
+  // scroll (paired with the header/footer `position: sticky` above) fixes that.
+  @include above('small') {
+    max-height: 90vh;
+    overflow-y: auto;
+  }
 
   @include below('small') {
     --border-radius: 0;


felipefialho (changes requested):
## Summary

This is a well-diagnosed fix: `ion-modal::part(content)` had no `max-height`, so tall content pushed the header/footer off-screen instead of scrolling. Capping it and pinning header/footer via `position: sticky` is the right approach, and scoping it to `above('small')` correctly leaves the existing mobile fullscreen modal untouched.

One issue should be fixed before merge:

- `background: var(--background)` on the new sticky header/footer isn't an Atomium token, it silently inherits Ionic's own `--background` custom prop with no fallback, which risks the sticky header/footer rendering transparent and letting content bleed through (see inline comments).

Also left two non-blocking suggestions (focus scrolling under the new sticky elements, and a dormant double-overflow risk in `.atom-modal__content`) and one praise comment.

**Positive points:**
- Root cause is clearly diagnosed and documented, including the `height: auto` vs percentage-height-on-auto-ancestor quirk.
- Correctly scoped to `above('small')` only, mobile behavior is untouched.
- The `90vh` magic number is paired with an explanatory comment per project convention.

packages/core/src/components/modal/modal.scss
```diff
@@ -38,13 +38,29 @@
     gap: var(--spacing-medium);
     justify-content: flex-end;
     padding: var(--spacing-large);
+
+    // Pins the footer while the content below scrolls (see ion-modal::part(content)).
+    @include above('small') {
+      background: var(--background);
```
felipefialho: [issue] `var(--background)` here isn't an Atomium design token: it doesn't exist in `packages/tokens/src/tokens.scss`. It resolves via inheritance to Ionic's own built-in `--background` custom prop on `<ion-modal>` (default `var(--ion-background-color, #fff)`), since `class="atom-modal"` is set directly on the `<ion-modal>` host and `.atom-modal` never defines `--background` itself. Every other component here that consumes `--background` (`.atom-chip`, `.is-outline-filled` in `button.scss`) defines it locally first before using it. With no fallback, if a consumer overrides `--background`/`--ion-background-color` for theming, or Ionic changes its default, the sticky footer could render with an unexpected or transparent background, letting scrolling content bleed through behind it. That defeats the purpose of this fix. Suggest an explicit fallback, e.g. `background: var(--background, var(--color-neutral-white));`, or defining `--background` locally on `.atom-modal`.

packages/core/src/components/modal/modal.scss
```diff
@@ -38,13 +38,29 @@
...
 
   &__header {
     color: var(--color-neutral-light-1);
     padding: var(--spacing-large);
     padding-right: var(--atom-modal-padding-header);
     position: relative;
+
+    // Pins the header while the content below scrolls (see ion-modal::part(content)).
+    @include above('small') {
+      background: var(--background);
```
felipefialho: [issue] Same `var(--background)` concern as the footer above: not an Atomium token, inherited from Ionic's `--background` prop with no fallback. Same fix applies here.

packages/core/src/components/modal/modal.scss
```diff
@@ -38,13 +38,29 @@
...
   }
 
   &__header {
     color: var(--color-neutral-light-1);
     padding: var(--spacing-large);
     padding-right: var(--atom-modal-padding-header);
     position: relative;
+
+    // Pins the header while the content below scrolls (see ion-modal::part(content)).
+    @include above('small') {
```
felipefialho, line 65: [suggestion] Now that `ion-modal::part(content)` is a real scroll container, a keyboard user tabbing to a focusable element near the top/bottom of the content can end up scrolled to a position visually covered by this sticky header/footer, since native focus `scrollIntoView` doesn't reserve space for sibling sticky elements. Consider `scroll-padding-top`/`scroll-padding-bottom` on `ion-modal::part(content)`, sized to header/footer height, to keep focused elements visible.

packages/core/src/components/modal/modal.scss
```diff
@@ -92,6 +108,14 @@ ion-modal::part(content) {
   --border-radius: 8px;
   height: auto;
 
+  // With no max-height, content taller than the viewport just overflows past both edges instead
+  // of scrolling internally, pushing the header/footer off-screen. Capping it and letting it
+  // scroll (paired with the header/footer `position: sticky` above) fixes that.
+  @include above('small') {
+    max-height: 90vh;
```
felipefialho, line 122: [suggestion] `.atom-modal__content` still has `height: 100%; overflow-y: auto` (unchanged). It stays inert today only because `max-height` alone doesn't give `ion-modal::part(content)` a definite height, so the percentage resolves to `auto`. That's an implicit invariant rather than an enforced one: if a future change gives this element an explicit `height` (e.g. to fix a Safari `max-height` quirk), `.atom-modal__content` would activate its own scrollbar nested inside this one. Worth making the single-scroll-container intent explicit, e.g. `.atom-modal__content { @include above('small') { overflow-y: visible; } }`, so it doesn't silently break later.

packages/core/src/components/modal/modal.scss
```diff
@@ -92,6 +108,14 @@ ion-modal::part(content) {
   --border-radius: 8px;
   height: auto;
 
+  // With no max-height, content taller than the viewport just overflows past both edges instead
```
felipefialho, line 118: [praise] Good use of an explanatory comment to justify the `90vh` magic number and cross-reference the paired `position: sticky` rules. Makes the reasoning traceable for whoever touches this next.

packages/core/src/components/modal/modal.scss
igorwessel: Good catch — fixed in 0e07ec79c: added a fallback, `background: var(--background, var(--color-neutral-white));`, matching the pattern already used in `tooltip.scss:21`.

packages/core/src/components/modal/modal.scss
igorwessel: Fixed alongside the footer one in 0e07ec79c — same fallback: `background: var(--background, var(--color-neutral-white));`.

packages/core/src/components/modal/modal.scss
igorwessel: Addressed in 0e07ec79c: added `scroll-padding-block: var(--spacing-xxxxlarge);` on `ion-modal::part(content)` so native focus scrolling reserves space for the sticky header/footer. It's an approximation sized for a single-line header/footer, but it's a solid improvement over reserving no space at all.

packages/core/src/components/modal/modal.scss
igorwessel: Addressed in 0e07ec79c: added `overflow-y: visible;` on `.atom-modal__content` under `above('small')` to make the single-scroll-container intent explicit, so it won't silently regress into a nested scrollbar later.

packages/core/src/components/modal/modal.scss
felipefialho: Confirmed at HEAD (`scroll-padding-block: var(--spacing-xxxxlarge);`). Good approximation, thanks!

packages/core/src/components/modal/modal.scss
felipefialho: Confirmed at HEAD (`overflow-y: visible` under `above('small')`). Thanks!

felipefialho (approved):
All previous issues resolved and verified at HEAD:

- `var(--background)` fallback added on both sticky header and footer
- `scroll-padding-block` added on `ion-modal::part(content)` for keyboard focus
- `overflow-y: visible` added on `.atom-modal__content` to keep a single scroll container

Great work!

dyhalmeida (approved):

mauriciomutte (approved):

brunosf (approved):

PR #863, opened 2026-07-13
fix(core): resolve hydrate to a browser no-op in client builds

igorwessel:
## Problem

Enabling SSR on the Vue/React output targets (`hydrateModule`, added in #853) made every generated proxy reference the Node-only hydrate app in its server-only branch:

```js
export const AtomModal = globalThis.window
  ? defineContainer('atom-modal', ...)                            // browser — never touches hydrate
  : defineStencilSSRComponent({ hydrateModule: import('@juntossomosmais/atomium/hydrate') })  // server only
```

Bundlers statically bundle **every** `import()`, regardless of which branch runs at runtime. So a **client-only** consumer (a Vite SPA, or Nuxt with `ssr: false`) ends up pulling `hydrate/index.mjs`, which does `import { Readable } from 'stream'`, and the browser build fails:

```
RollupError: node_modules/@juntossomosmais/atomium/hydrate/index.mjs (1:9):
"Readable" is not exported by "__vite-browser-external"
```

## Fix

Add a `browser` export condition on `./hydrate` pointing to a generated no-op stub. Vite/webpack/Rollup/esbuild resolve `browser` **only in client builds** and ignore it in server builds:

- **client builds** (SPA / `ssr: false`, and the client half of an SSR app) → resolve the empty stub → no `stream` import → build passes.
- **server builds** (Nitro/Node) → keep `hydrate/index.mjs` → SSR still works.

The stub mirrors the hydrate public API as `serverOnly` functions that throw, so any accidental browser-side use surfaces loudly. It's never invoked in a browser (its branch only runs when `globalThis.window` is falsy).

### Changes
- `packages/core/scripts/write-hydrate-browser-stub.mjs` — generates `hydrate/index.browser.mjs` post-build.
- `packages/core/package.json` — `browser` condition on `./hydrate` + wire the script into `build`.

## Verification
- `nx build @juntossomosmais/atomium` ✅ — build finished; `browser` condition survives the stencil build (not rewritten).
- `ssr-smoke.mjs` (renders all components via the **real** hydrate in Node) ✅ 25/25 — server SSR intact.

## Notes
Stencil's SSR docs assume a full SSR framework and don't cover keeping the Node hydrate out of a client-only bundle. The `browser` export condition is the standard package-level convention for this. Worth reporting the gap upstream in `stencil-ds-output-targets`.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -28,7 +28,7 @@
     "core.css"
   ],
   "scripts": {
-    "build": "stencil build && npx copyfiles --flat ./dist/core/core.css ./ && node -e \"require('fs').writeFileSync('dist/components/package.json', JSON.stringify({ type: 'module' }, null, 2) + '\\n')\" && node scripts/ssr-smoke.mjs",
+    "build": "stencil build && npx copyfiles --flat ./dist/core/core.css ./ && node -e \"require('fs').writeFileSync('dist/components/package.json', JSON.stringify({ type: 'module' }, null, 2) + '\\n')\" && node scripts/write-hydrate-browser-stub.mjs && node scripts/ssr-smoke.mjs",
     "start": "stencil build --watch",
     "test": "stencil test --spec",
     "test:ci": "stencil test --spec --coverage --no-cache --updateSnapshot --ci",
@@ -51,6 +51,7 @@
     "./dist/*": "./dist/*",
     "./hydrate": {
       "types": "./hydrate/index.d.ts",
+      "browser": "./hydrate/index.browser.mjs",
       "import": "./hydrate/index.mjs",
       "require": "./hydrate/index.js"
     },
--- a/packages/core/scripts/write-hydrate-browser-stub.mjs
+++ b/packages/core/scripts/write-hydrate-browser-stub.mjs
@@ -14,31 +14,25 @@
  * These functions are never invoked in a browser (the branch that uses them only
  * runs when `globalThis.window` is falsy), so they throw to surface any misuse.
  *
+ * Export names are derived from the real hydrate module (not hand-duplicated),
+ * so this stub can never drift from whatever Stencil generates.
+ *
  * Runs as part of `npm run build`, after `stencil build` has generated `hydrate/`.
  */
-import { writeFileSync } from 'fs'
 import { fileURLToPath } from 'url'
 import { dirname, resolve } from 'path'
 
-const target = resolve(
+import { writeHydrateBrowserStub } from './hydrate-browser-stub.mjs'
+import * as realHydrateModule from '../hydrate/index.mjs'
+
+const exportNames = Object.keys(realHydrateModule).sort()
+const targetPath = resolve(
   dirname(fileURLToPath(import.meta.url)),
   '../hydrate/index.browser.mjs',
 )
 
-const stub = `/* auto-generated — browser no-op for the Node-only hydrate app. See scripts/write-hydrate-browser-stub.mjs */
-const serverOnly = () => {
-  throw new Error('@juntossomosmais/atomium/hydrate is server-only and cannot run in the browser')
-}
-export const renderToString = serverOnly
-export const streamToString = serverOnly
-export const hydrateDocument = serverOnly
-export const createWindowFromHtml = serverOnly
-export const serializeDocumentToString = serverOnly
-export const serializeProperty = serverOnly
-export const deserializeProperty = serverOnly
-export const setTagTransformer = serverOnly
-export const transformTag = serverOnly
-`
+writeHydrateBrowserStub(exportNames, targetPath)
 
-writeFileSync(target, stub)
-console.log('Wrote hydrate/index.browser.mjs (browser no-op)')
+console.log(
+  `Wrote hydrate/index.browser.mjs (browser no-op) mirroring ${exportNames.length} export(s): ${exportNames.join(', ')}`,
+)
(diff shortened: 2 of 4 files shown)


felipefialho (changes requested):
Solid, well-scoped fix for a real bundler issue (browser export condition is exactly the right lever here, and the PR description/verification notes are excellent). A few things should be addressed before merge:

**Summary**
1. [issue] The stub's 9 exported function names are hand-duplicated from the real `hydrate` module instead of derived/verified against it — silent API drift risk.
2. [issue] No test file exists for the new `write-hydrate-browser-stub.mjs` script, which contains real logic (file writing, path resolution) and isn't exempt under the project's test-coverage convention.
3. [issue] No automated regression test simulates a client-only bundler resolving the `browser` condition to confirm `stream` is actually excluded — the exact scenario this PR fixes is only verified manually.

None of these are security concerns (reviewed and clean), but they're maintainability/coverage gaps worth closing before merge.

packages/core/package.json
```diff
@@ -51,6 +51,7 @@
     "./dist/*": "./dist/*",
     "./hydrate": {
       "types": "./hydrate/index.d.ts",
+      "browser": "./hydrate/index.browser.mjs",
       "import": "./hydrate/index.mjs",
```
felipefialho, line 55: [praise] This is the right fix — the `browser` export condition is the standard convention bundlers use to pick a client-safe entry point, and it correctly leaves the `import`/`require` conditions untouched for SSR. Clean, minimal, targeted change.

packages/core/scripts/write-hydrate-browser-stub.mjs
```diff
@@ -0,0 +1,44 @@
...
+const target = resolve(
+  dirname(fileURLToPath(import.meta.url)),
+  '../hydrate/index.browser.mjs',
+)
+
+const stub = `/* auto-generated — browser no-op for the Node-only hydrate app. See scripts/write-hydrate-browser-stub.mjs */
+const serverOnly = () => {
+  throw new Error('@juntossomosmais/atomium/hydrate is server-only and cannot run in the browser')
+}
+export const renderToString = serverOnly
```
felipefialho: [issue] These 9 export names are hand-duplicated from the real `hydrate/index.mjs` (or `index.d.ts`) instead of derived from it. If Stencil's `dist-hydrate-script` output target adds/renames/removes an export in the future, this stub won't follow automatically — a client build would silently get `undefined` for a missing export instead of a build-time signal, which is the exact class of failure this PR is fixing (a divergence that only surfaces at runtime in client-only builds). Consider generating this list programmatically from `hydrate/index.d.ts`'s exports, or adding a build-time assertion that fails CI if the stub's export set and the real module's export set diverge.

packages/core/scripts/write-hydrate-browser-stub.mjs
```diff
@@ -0,0 +1,44 @@
...
+export const hydrateDocument = serverOnly
+export const createWindowFromHtml = serverOnly
+export const serializeDocumentToString = serverOnly
+export const serializeProperty = serverOnly
+export const deserializeProperty = serverOnly
+export const setTagTransformer = serverOnly
+export const transformTag = serverOnly
+`
+
+writeFileSync(target, stub)
```
felipefialho: [issue] This new script has file-writing/path-resolution logic and isn't in the exempt categories (pure constants, type-only, static data) from the project's test convention, but there's no corresponding test file. Additionally — and more importantly — there's no automated regression test for the bug this PR actually fixes: nothing simulates a client-only bundler (Vite/webpack) resolving the `browser` condition and asserts `stream`/other Node builtins are absent from the resulting graph. The current verification (`nx build` + `ssr-smoke.mjs`) only exercises the SSR path, not the browser-condition path. Worth adding at least a minimal fixture test for the client-build resolution, since a future refactor of the export target or package.json exports map could silently reintroduce this exact bug with no CI signal.

packages/core/scripts/write-hydrate-browser-stub.mjs
igorwessel: Fixed in 2dee47d — export names are now derived from `Object.keys()` of the real, built hydrate module (`import * as realHydrateModule from '../hydrate/index.mjs'`) rather than hand-written. The stub-generation logic moved to a tested helper (`hydrate-browser-stub.mjs`), so if Stencil ever adds/renames/removes a hydrate export, the stub picks it up automatically on the next build — no manual list to keep in sync.

packages/core/scripts/write-hydrate-browser-stub.mjs
igorwessel: Fixed in 2dee47d:
- Added `packages/core/scripts/hydrate-browser-stub.spec.ts`, covering the pure stub-generation logic (`buildHydrateBrowserStub`) and the file-write call (`writeHydrateBrowserStub`, now DI-friendly — takes a `writeFile` fn — so it's testable without mocking the `fs` module).
- On the regression-test point: I couldn't wire a true "run a bundler against the built package" test into `test`/`test:ci`, because `hydrate/` is gitignored and generated only by `npm run build`, and CI's Sonar workflow runs `test:ci`/`test:ci-merge` without a preceding build — so a spec depending on build output would be flaky/absent in CI. Instead I added two deterministic tests that together pin down the actual fix mechanism without needing build artifacts:
  1. a test asserting the generated stub contains **zero** `import`/`require` statements — this is precisely why no Node builtin (stream included) can ever be pulled into a client bundle through it, regardless of what the real hydrate module does;
  2. a test reading the committed `package.json` asserting `exports['./hydrate'].browser` exists and points at a file distinct from `.import`/`.require` — so a future edit that drops the condition fails CI immediately.

Happy to also add a build-time (non-jest) check if you'd prefer belt-and-suspenders — let me know.

felipefialho (approved):
Both prior issues resolved: export names are now derived from `Object.keys(realHydrateModule)` instead of hand-duplicated, and solid unit tests were added for the stub generator plus a package.json export-map assertion. Great work, thanks for the thorough follow-up.

brunosf (approved):

luiz-ortega (approved):

PR #867, opened 2026-07-28
fix(core): dismiss body-level ion-popover on scroll without ion-app

dyhalmeida:
[Task](https://juntossomosmais.monday.com/boards/5337789717/pulses/12649333296)

## Summary

Fixes `atom-select`'s options popover staying pinned at fixed viewport coordinates and floating over unrelated page content while the user scrolls, instead of following its trigger or closing.

**Root cause**: `atom-select` renders `<ion-select interface="popover">`. On present, Ionic teleports the popover's content into `<ion-app>` — that's the element responsible for repositioning/dismissing overlays as the page scrolls. Any host that doesn't mount an `<ion-app>` root (e.g. a plain Next.js/React app consuming Atomium components directly, without the full Ionic framework shell) falls back to Ionic's default teleport target, `document.body`, which has no such wiring. The popover then just stays wherever it opened.

This is not specific to `atom-select`'s SSR/hydration setup — it reproduces identically whether the component is server-rendered or fully client-hydrated, and regardless of how it's consumed (direct JSX or wrapped in `styled()`). It's a gap in how `interface="popover"` behaves outside a full Ionic app shell.

**Fix**: `packages/core/src/global/global.ts` now registers a `window` scroll listener as a module-level side effect (so it also runs in the `dist-custom-elements`/React output target, which never invokes the exported `globalScript()` function — same pattern already used there for `forceMdMode`). The listener dismisses any `ion-popover` that's a direct child of `body`, and only when no `<ion-app>` is present, so it never interferes with a real Ionic app's own overlay handling.

Verified against the two `AtomSelect` consumption patterns in a downstream app (direct JSX and `styled(AtomSelect)`) — scrolling with the popover open now closes it instead of leaving it floating.

`pnpm test`: 246/246 passing, 29 suites.

## Evidences

https://github.com/user-attachments/assets/c8feacdb-b065-4f6f-bc89-b0ca070cff53

### DIFF
--- a/packages/core/src/global/global.ts
+++ b/packages/core/src/global/global.ts
@@ -3,6 +3,12 @@ import '@ionic/core'
 // Polyfill need to make form.requestSubmit work in Safari < 16.
 import '../polyfills/form-request-submit.js'
 
+declare global {
+  interface Window {
+    __atomiumDismissPopoversOnScroll?: boolean
+  }
+}
+
 // Force Ionic's "md" mode. The dist-custom-elements output consumed by the
 // React wrappers never runs global scripts, so the mode resolver must be set
 // as a module side-effect here, otherwise Ionic-backed components render
@@ -19,6 +25,35 @@ if (typeof setMode === 'function') {
   forceMdMode()
 } else {
   queueMicrotask(forceMdMode)
+}
+
+// ion-select's popover interface (and any other Ionic overlay using
+// `interface="popover"`) teleports its content to `ion-app` on present so the
+// app root can reposition/dismiss it as the page scrolls. Hosts that don't
+// mount an `<ion-app>` (e.g. a Next.js app consuming Atomium components
+// directly) get the fallback teleport target instead: `document.body`
+// (see Ionic's `getAppRoot`). Nothing then dismisses the overlay on scroll,
+// so it stays pinned at the viewport coordinates where it opened and floats
+// over unrelated content as the page scrolls underneath it. Since there is
+// no `ion-app` to patch, dismiss any body-level popover here instead — this
+// module runs as a side effect on import, unlike `globalScript` below, which
+// the dist-custom-elements/React output target never invokes (see comment
+// above). Guarded so re-importing this module (multiple lazy chunks in the
+// legacy `dist` runtime) only ever registers one listener.
+if (typeof window !== 'undefined' && !window.__atomiumDismissPopoversOnScroll) {
+  window.__atomiumDismissPopoversOnScroll = true
+
+  window.addEventListener(
+    'scroll',
+    () => {
+      if (document.querySelector('ion-app')) return
+
+      document.body
+        .querySelectorAll(':scope > ion-popover')
+        .forEach((popover) => (popover as HTMLIonPopoverElement).dismiss())
+    },
+    { passive: true }
+  )
 }
 
 export default function globalScript() {
--- a/packages/core/src/global/global.ts
+++ b/packages/core/src/global/global.ts
@@ -39,21 +39,107 @@ if (typeof setMode === 'function') {
 // module runs as a side effect on import, unlike `globalScript` below, which
 // the dist-custom-elements/React output target never invokes (see comment
 // above). Guarded so re-importing this module (multiple lazy chunks in the
-// legacy `dist` runtime) only ever registers one listener.
+// legacy `dist` runtime) only ever registers one set of listeners.
 if (typeof window !== 'undefined' && !window.__atomiumDismissPopoversOnScroll) {
   window.__atomiumDismissPopoversOnScroll = true
 
-  window.addEventListener(
-    'scroll',
-    () => {
-      if (document.querySelector('ion-app')) return
+  // Only Atomium's own popovers are dismissed, so a host pairing Atomium with
+  // another Ionic-based library keeps control of its own overlays. The marker
+  // class is applied through `interfaceOptions.cssClass` (see select.tsx).
+  const ATOMIUM_POPOVER = 'ion-popover.atomium-popover'
+  const BODY_POPOVERS = `:scope > ${ATOMIUM_POPOVER}`
 
-      document.body
-        .querySelectorAll(':scope > ion-popover')
-        .forEach((popover) => (popover as HTMLIonPopoverElement).dismiss())
-    },
-    { passive: true }
-  )
+  // `capture: true` so scroll events from descendant `overflow:auto`
+  // containers are observed too — `scroll` doesn't bubble, but a capture
+  // listener on `window` still sees it during the capture phase.
+  const SCROLL_OPTIONS: AddEventListenerOptions = {
+    capture: true,
+    passive: true,
+  }
+
+  let scheduled = false
+  let lastScrollTarget: EventTarget | null = null
+  let isListening = false
+
+  const getBodyPopovers = () =>
+    document.body.querySelectorAll<HTMLIonPopoverElement>(BODY_POPOVERS)
+
+  // Duck-typed rather than `instanceof Element`: an event target can originate
+  // from another realm (an iframe's document), where `instanceof` is false for
+  // a genuine element. `window`/`document` targets correctly yield null.
+  const asElement = (target: EventTarget | null): Element | null =>
+    typeof (target as Element | null)?.matches === 'function'
+      ? (target as Element)
+      : null
+
+  // rAF-throttled so the DOM queries below run at most once per frame
+  // instead of on every `scroll` event fired during a gesture.
+  const dismissBodyPopoversOnScroll = (): void => {
+    scheduled = false
+
+    // Read and release: holding the reference past this frame would keep a
+    // detached node alive for the lifetime of the page.
+    const scrollTarget = lastScrollTarget
+
+    lastScrollTarget = null
+
+    if (document.querySelector('ion-app')) return
+
+    // A scroll whose target lives inside the popover itself (e.g. paging
+    // through a long select's option list) is the intended interaction, not
+    // a page scroll — dismissing here would close the list mid-scroll.
+    if (asElement(scrollTarget)?.closest(ATOMIUM_POPOVER)) return
+
+    getBodyPopovers().forEach((popover) => popover.dismiss().catch(() => {}))
+  }
+
+  const handleScroll = (event: Event): void => {
+    lastScrollTarget = event.target
+
+    if (scheduled) return
+
+    scheduled = true
+    requestAnimationFrame(dismissBodyPopoversOnScroll)
+  }
+
+  const startListening = (): void => {
+    if (isListening) return
+
+    isListening = true
+    window.addEventListener('scroll', handleScroll, SCROLL_OPTIONS)
+  }
+
+  const stopListening = (): void => {
+    if (!isListening) return
+
+    isListening = false
+    scheduled = false
+    lastScrollTarget = null
+    window.removeEventListener('scroll', handleScroll, SCROLL_OPTIONS)
+  }
+
+  const isAtomiumPopover = (target: EventTarget | null): boolean =>
+    asElement(target)?.matches(ATOMIUM_POPOVER) === true
+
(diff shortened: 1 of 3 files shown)


felipefialho (changes requested):
Thanks for the clear root-cause writeup and the repro video — the diagnosis (Ionic's popover teleport falling back to `document.body` outside an `<ion-app>` shell) is solid and the guard against double-registration in the lazy-chunk runtime is a nice touch.

**Summary of blocking issues:**
1. The `scroll` handler runs two unthrottled DOM queries (`querySelector('ion-app')` + `querySelectorAll(':scope > ion-popover')`) on every single scroll event, for the lifetime of every page that imports Atomium — a real INP/main-thread cost paid even when no popover is ever opened.
2. `scroll` events don't bubble, so a non-capturing `window` listener never fires when the page scrolls via a nested `overflow: auto` container (a common layout, especially in the Next.js apps this fix targets) — meaning the exact bug this PR fixes may still reproduce in that setup.
3. This introduces the first non-trivial branching logic in `global.ts` and it ships with zero test coverage, even though the repo already has an established, easy-to-follow pattern for this kind of DOM/global mocking (see `packages/core/src/utils/screens.spec.ts`).

See inline comments for details and suggested fixes.

packages/core/src/global/global.ts
```diff
@@ -21,6 +27,35 @@ if (typeof setMode === 'function') {
...
 
+// ion-select's popover interface (and any other Ionic overlay using
+// `interface="popover"`) teleports its content to `ion-app` on present so the
+// app root can reposition/dismiss it as the page scrolls. Hosts that don't
+// mount an `<ion-app>` (e.g. a Next.js app consuming Atomium components
+// directly) get the fallback teleport target instead: `document.body`
+// (see Ionic's `getAppRoot`). Nothing then dismisses the overlay on scroll,
+// so it stays pinned at the viewport coordinates where it opened and floats
+// over unrelated content as the page scrolls underneath it. Since there is
+// no `ion-app` to patch, dismiss any body-level popover here instead — this
```
felipefialho, line 38: [issue] Two DOM queries run synchronously on every `scroll` event (`document.querySelector('ion-app')` here, and `querySelectorAll(':scope > ion-popover')` below), with no throttling/debouncing/rAF gating. `scroll` can fire dozens of times per second during a gesture, and this cost is paid by every page that imports Atomium for the lifetime of the page — even pages that never open a popover.

Suggestion: only register this listener while a body-level popover is actually presented (e.g. hook into `ionPopoverDidPresent`/`ionPopoverDidDismiss`), or at minimum rAF-throttle the handler body so the DOM work runs at most once per frame.

packages/core/src/global/global.ts
```diff
@@ -21,6 +27,35 @@ if (typeof setMode === 'function') {
...
+// module runs as a side effect on import, unlike `globalScript` below, which
```
felipefialho, line 39: [issue] `scroll` events don't bubble — a listener on `window` without `capture: true` only fires when the document itself scrolls, not when the user scrolls an inner `overflow: auto` container. That's a common layout in exactly the kind of app this fix targets (a Next.js host without a full Ionic `<ion-app>` shell often has its own fixed layout with an inner scroll region). In that case the popover would still stay pinned as the page scrolls underneath it, which is the bug this PR sets out to fix.

Consider adding `capture: true` so scroll events from descendant scrollable elements are observed too (and guard against dismissing a popover because of scroll *within* the popover itself, if that becomes relevant).

packages/core/src/global/global.ts
```diff
@@ -21,6 +27,35 @@ if (typeof setMode === 'function') {
...
+// the dist-custom-elements/React output target never invokes (see comment
+// above). Guarded so re-importing this module (multiple lazy chunks in the
+// legacy `dist` runtime) only ever registers one listener.
+if (typeof window !== 'undefined' && !window.__atomiumDismissPopoversOnScroll) {
+  window.__atomiumDismissPopoversOnScroll = true
```
felipefialho, line 44: [suggestion] This dismisses *every* direct `ion-popover` child of `body`, not just ones opened by Atomium components. If a host page uses another Ionic-based library or its own `ion-popover` alongside Atomium, this would silently close overlays this library doesn't own. Worth scoping the selector to Atomium's own popovers (e.g. a marker attribute set via `interfaceOptions` in `select.tsx`/`datetime.tsx`) if that cross-boundary side effect is a concern.

packages/core/src/global/global.ts
```diff
@@ -21,6 +27,35 @@ if (typeof setMode === 'function') {
...
+
```
felipefialho, line 45: [suggestion] Two small polish items here:
1. `(popover as HTMLIonPopoverElement)` is an unguarded `as` cast — using a typed generic on the query instead avoids it entirely: `querySelectorAll<HTMLIonPopoverElement>(':scope > ion-popover')`.
2. `dismiss()` returns a `Promise<boolean>` and is called fire-and-forget inside `forEach`. If it rejects (e.g. overlay mid-teardown on a rapid-fire scroll before any throttling is added), it becomes an unhandled promise rejection. Consider `popover.dismiss().catch(() => {})`.

packages/core/src/global/global.ts
```diff
@@ -21,6 +27,35 @@ if (typeof setMode === 'function') {
   queueMicrotask(forceMdMode)
 }
 
+// ion-select's popover interface (and any other Ionic overlay using
+// `interface="popover"`) teleports its content to `ion-app` on present so the
+// app root can reposition/dismiss it as the page scrolls. Hosts that don't
+// mount an `<ion-app>` (e.g. a Next.js app consuming Atomium components
+// directly) get the fallback teleport target instead: `document.body`
+// (see Ionic's `getAppRoot`). Nothing then dismisses the overlay on scroll,
```
felipefialho, line 35: [issue] This is the first non-trivial branching logic added to `global.ts` (the guard, the `ion-app` presence check, and the dismiss-on-scroll behavior), and it has zero automated test coverage. The manual video verification confirms the happy path, but won't catch regressions in the guard (e.g. accidentally dismissing popovers even when `ion-app` is present, or double-registering the listener across lazy chunks).

The repo already has an established, easy-to-follow pattern for exactly this kind of DOM/global mocking — see `packages/core/src/utils/screens.spec.ts` (uses `Object.defineProperty`/`jest.fn()` to stub browser globals). A colocated `global.spec.ts` covering: (1) dismiss called when no `ion-app` + a popover present, (2) dismiss NOT called when `ion-app` present, (3) listener registered only once across repeated imports, would lock in the behavior this PR is fixing.

packages/core/src/global/global.ts
```diff
@@ -21,6 +27,35 @@ if (typeof setMode === 'function') {
...
+// module runs as a side effect on import, unlike `globalScript` below, which
+// the dist-custom-elements/React output target never invokes (see comment
+// above). Guarded so re-importing this module (multiple lazy chunks in the
+// legacy `dist` runtime) only ever registers one listener.
+if (typeof window !== 'undefined' && !window.__atomiumDismissPopoversOnScroll) {
+  window.__atomiumDismissPopoversOnScroll = true
+
+  window.addEventListener(
+    'scroll',
+    () => {
```
felipefialho: [praise] Good attention to detail guarding against duplicate listener registration when this module is re-imported across multiple lazy chunks in the `dist` runtime — and reusing the same module-side-effect pattern already established by `forceMdMode` above (with a clear comment explaining why it has to work this way for the `dist-custom-elements`/React output target) keeps the file consistent.

packages/core/src/global/global.ts
dyhalmeida: done a4ef561f0 3f92ad711

Went with your first option rather than the rAF-only floor: the scroll listener is now attached on `ionPopoverDidPresent` and detached once the last Atomium popover closes, so a page that never opens one pays nothing for the lifetime of the page. The handler is rAF-throttled on top of that.

Worth flagging from Ionic's source: `didDismiss` is emitted at `overlays.js:613` but `el.remove()` only at line 638, so the element is still attached when the event fires — a naive "any popovers left?" check would never detach. That check is deferred one microtask. I also chose re-querying the DOM over counting present/dismiss pairs, since a drifting counter would leave the listener attached forever, whereas re-querying self-heals.

packages/core/src/global/global.ts
dyhalmeida: done a4ef561f0 3f92ad711

Added `capture: true`. Also implemented the parenthetical: a scroll whose target lives inside the popover no longer dismisses it, so paging through a long option list doesn't close the list mid-scroll.

Note the tests can't exercise capture *behaviour* — Stencil's `@stencil/core/testing` preset replaces jsdom with mock-doc, which doesn't implement capture-phase propagation at all (a non-bubbling event on a descendant never reaches a `window` capture listener). So that test asserts the registered listener options instead, and the intra-popover test dispatches with `bubbles: true` to route an event with the right target to the handler. Both are commented in the spec so the accommodation is explicit rather than looking like a claim that `scroll` bubbles.

packages/core/src/global/global.ts
dyhalmeida: done 101fba001 a4ef561f0

Agreed the cross-boundary side effect was worth avoiding. `atom-select` now tags its popover via `interfaceOptions.cssClass` and the query is scoped to `:scope > ion-popover.atomium-popover`, so overlays owned by another library are left alone. `atom-select` exposes no `interfaceOptions` prop, so consumers can't accidentally wipe the marker.

packages/core/src/global/global.ts
dyhalmeida: done a4ef561f0

Both applied: the query is now `querySelectorAll<HTMLIonPopoverElement>(...)` so the cast is gone, and `dismiss()` gets a `.catch(() => {})`.

While in there I also dropped an `instanceof Element` check I'd added for the scroll-target guard in favour of a duck-typed helper — `instanceof` is false for a genuine element originating from another realm (an iframe's document), which is the same class of unguarded-assumption problem as the cast.

packages/core/src/global/global.ts
dyhalmeida: done 3f92ad711

gabrielduete (approved):

packages/core/src/global/global.ts
felipefialho: Confirmed in code, the listener is now attached on `ionPopoverDidPresent` / detached on `ionPopoverDidDismiss`, and rAF-throttled. Great write-up on the microtask timing quirk too. Thanks!

packages/core/src/global/global.ts
felipefialho: Confirmed, `capture: true` is set and the intra-popover scroll guard works as described. Good call documenting the mock-doc limitation in the spec instead of leaving it implicit. Thanks!

packages/core/src/global/global.ts
felipefialho: Confirmed, the selector is scoped to `ion-popover.atomium-popover` and `select.tsx` applies the marker via `interfaceOptions.cssClass`. Thanks!

packages/core/src/global/global.ts
felipefialho: Confirmed, both applied: typed generic query and `.catch(() => {})` on dismiss. Good catch on swapping the `instanceof` check too. Thanks!

packages/core/src/global/global.ts
felipefialho: Confirmed, the new `global.spec.ts` covers the guard, the ion-app presence check, and single-registration across imports. Thanks!

felipefialho (approved):
All previously raised issues verified as fixed in the current code: rAF-throttled scroll listener attached/detached with popover lifecycle, `capture: true` with an intra-popover scroll guard, popover queries scoped to `.atomium-popover`, cast removed with a caught `dismiss()`, and solid new test coverage in `global.spec.ts`. Great work, thanks for addressing everything thoroughly!

3rdepaula (approved):

igorwessel (approved):

